PR #10, opened 2021-08-30
INFRA-6777 - Update testing jobs to pass if when there are no tests

dbraley:
All of our testing jobs end with a required analysis phase. Unfortunately analysis fails if there are no testing reports to analyze. This can be a good thing, since there are legitimately times where test reports might fail to be generated or saved, which should cause the job to fail. However, there are other times we have entered the testing phase, but have no tests to run - either they've been marked as already run elsewhere in the case of e2e tests, or there are just no tests to run for whatever has changed in the case of unit tests. In these scenarios, there are no test reports generated, and that fails analysis, when in reality there is no problem and the stage should complete correctly.

Here we detect this prior to running the tests, then set a variable to let our analysis phase later know that it can be skipped as it will not find any reports, thus passing the testing job (assuming there was no other failure in the interim).

# Test Plan:
## Unit Tests
1. See [webapp-test-113728](https://jenkins.khanacademy.org/job/deploy/job/webapp-test/113728/consoleText), a webapp test job that was run with the code here* that had multiple unit tests to run, and does correctly report that it enters and completes the `Analyzing results` stage, and that the job completes successfully.
2. See [webapp-test-113727](https://jenkins.khanacademy.org/job/deploy/job/webapp-test/113727/consoleText), a webapp test job that was run with the code here* that had zero unit tests to run, and logs at the end of the run that it is `Skipping Analysis - No tests run`. The job completed successfully.
3. To run this yourself, create a branch with a file deleted, such as a local dev script (or use [this](https://github.com/khan/webapp/tree/INFRA-6777-dummy-webapp-branch) branch). Have sun run the test job on the branch. If prior to merging of this code, rerun the job with this branches `webapp-test.groovy`, as outlined [here](https://khanacademy.atlassian.net/wiki/spaces/INFRA/pages/546931008/Introduction+to+Jenkins+Administration#Test-by-Replay). Observe the test passes.

## E2E Tests
1. See [e2e-test 31199](https://jenkins.khanacademy.org/job/deploy/job/e2e-test/31199/consoleText), an end to end test job run with the code here* that had one e2e test to run, and does correct log at the end of the file that it enters and completes the `Analyzing results` stage, and that the job completes successfully.
2. See [e2e-test 31202](https://jenkins.khanacademy.org/job/deploy/job/e2e-test/31202/consoleText), an end to end test job run with the code here* that had no e2e tests to run, and logs at the end of the run that it is `Skipping Analysis - No tests run`. The job completed successfully.
3. To run this yourself, find any run of the e2e tests, the click rebuild and modify the `TESTS_TO_RUN` field and the `SKIP_TESTS` fields to have the same value of a valid test (technically, you can include more skip tests, but it’s not necessary). For the sake of not being a nuisance, you should point the slack channel at your slack name or some other low traffic channel, clear the slack thread, lower the number of workers (and make sure they’re not first-in-queue), and rename the test to something that indicates what you’re doing. Then run that, and you’ll observe the tail of the log indicates `ERROR: None of the test reports contained any result`. Then click **Replay** and replace the code with this version of the `e2e-test.groovy` and notice that the run passes by skipping the analysis phase.

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -293,6 +293,28 @@ def runTestServer() {
          runSmokeTestsArgs += ["--skip-tests=${params.SKIP_TESTS}"];
       }
 
+      // Determine if we actually need to run any tests at all
+      // tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
+      tests = exec.outputOf(["testing/runtests_server.py", "-n", runSmokeTestsArgs, "."].flatten());
+
+      // The runtests_server.py script with -n outputs all tests to run of 
+      // various types. We only need to worry about smoke tests, which are 
+      // also the last section, so grab everything after the header. If it's 
+      // empty, there are no tests to run.
+      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 12)
+      // An empty line indicates the end of the section (if found)
+      emptyLineIndex = e2eTests.lastIndexOf('\n\n')
+      if (emptyLineIndex >= 0) {
+         e2eTests = e2eTests.substring(0, emptyLineIndex).trim()
+      }
+
+      echo("e2eTests: ${e2eTests}")
+      if (e2eTests.isAllWhitespace()) {
+         echo("No E2E Tests to run!")
+         skipTestAnalysis = true
+         throw new TestsAreDone();
+      }
+
       // This gets our 10.x.x.x IP address.
       def serverIP = exec.outputOf(["ip", "route", "get", "10.1.1.1"]).split()[6];
       // This unblocks the test-workers to let them know they can connect
@@ -304,19 +326,6 @@ def runTestServer() {
       // runtests_server.py writes to this directory.  Make sure it's clean
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
-
-      tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
-      // The runtests_server.py script with -n outputs all tests to run of 
-      // various types. We only need to worry about smoke tests, which are 
-      // also the last section, so grab everything after the header. If it's 
-      // empty, there are no tests to run.
-      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 13).trim()
-      echo("e2eTests: ${e2eTests}")
-      if (e2eTests.isAllWhitespace()) {
-         echo("No E2E Tests to run!")
-         skipTestAnalysis = true
-         return
-      }
 
       // START THE SERVER!  Note this blocks.  It will auto-exit when
       // it's done serving all the tests.
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -208,6 +208,9 @@ WORKER_TYPE = (params.USE_FIRSTINQUEUE_WORKERS
                ? 'ka-firstinqueue-ec2' : 'ka-test-ec2');
 
 
+// Gloabally scoped to allow the test runner to set this and allow us to skip analysis as well
+skipTestAnalysis = false
+
 // Run body, set the red circle on the flow-pipeline stage if it fails,
 // but do not fail the overall build.  Re-raises "interrupt" exceptions,
 // either due to a failFast or due to a user interrupt.
@@ -301,6 +304,19 @@ def runTestServer() {
       // runtests_server.py writes to this directory.  Make sure it's clean
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
+
+      tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
+      // The runtests_server.py script with -n outputs all tests to run of 
+      // various types. We only need to worry about smoke tests, which are 
+      // also the last section, so grab everything after the header. If it's 
+      // empty, there are no tests to run.
+      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 13).trim()
+      echo("e2eTests: ${e2eTests}")
+      if (e2eTests.isAllWhitespace()) {
+         echo("No E2E Tests to run!")
+         skipTestAnalysis = true
+         return
+      }
 
       // START THE SERVER!  Note this blocks.  It will auto-exit when
       // it's done serving all the tests.
@@ -555,6 +571,13 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
             runTests();
          }
       } finally {
+         // If we determined there were no tests to run, we should skip 
+         // analysis since it fails if there are no test results.
+         if (skipTestAnalysis) {
+            echo("Skipping Analysis - No tests run")
+            return
+         }
+
          // We want to analyze results even if -- especially if --
          // there were failures; hence we're in the `finally`.
          stage("Analyzing results") {
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -294,14 +294,17 @@ def runTestServer() {
       }
 
       // Determine if we actually need to run any tests at all
-      // tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
-      tests = exec.outputOf(["testing/runtests_server.py", "-n", runSmokeTestsArgs, "."].flatten());
+      // TODO(dbraley): This process takes a few seconds to run, and is done 
+      //  again by the actual server run a bit later in this method. We should
+      //  make this faster, or at least able to reuse the test list we've 
+      //  already calculated.
+      tests = exec.outputOf(["testing/runtests_server.py", "-n"] + runSmokeTestsArgs + ["."]);
 
       // The runtests_server.py script with -n outputs all tests to run of 
       // various types. We only need to worry about smoke tests, which are 
       // also the last section, so grab everything after the header. If it's 
       // empty, there are no tests to run.
-      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 12)
+      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + "SMOKE TESTS:".length())
       // An empty line indicates the end of the section (if found)
       emptyLineIndex = e2eTests.lastIndexOf('\n\n')
       if (emptyLineIndex >= 0) {
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -262,6 +262,14 @@ def runTestServer() {
                           "--lintfile=../files_to_lint.txt",
                         ];
 
+      // Determine if we actually need to run any tests at all
+      tests = exec.outputOf(["cat", "../files_to_test.txt"]);
+      if (tests.isAllWhitespace()) {
+         echo("No Unit Tests to run!")
+         skipTestAnalysis = true
+         throw new TestsAreDone();
+      }
+
       // This gets our 10.x.x.x IP address.
       def serverIP = exec.outputOf(["ip", "route", "get", "10.1.1.1"]).split()[6];
       // This unblocks the test-workers to let them know they can connect
@@ -279,13 +287,6 @@ def runTestServer() {
       // runtests_server.py writes to this directory.  Make sure it's clean
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
-
-      tests = sh(script: "cat ../files_to_test.txt", returnStdout: true);
-      if (tests.isAllWhitespace()) {
-         echo("No Unit Tests to run!")
-         skipTestAnalysis = true
-         return
-      }
 
       // START THE SERVER!  Note this blocks.  It will auto-exit when
       // it's done serving all the tests.
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -179,6 +179,8 @@ TESTS = [
     [cmd: "testing/go_test_client.sh -j1 <server>", done: false],
 ];
 
+// Gloabally scoped to allow the test runner to set this and allow us to skip analysis as well
+skipTestAnalysis = false
 
 // Run body, set the red circle on the flow-pipeline stage if it fails,
 // but do not fail the overall build.  Re-raises "interrupt" exceptions,
@@ -277,6 +279,13 @@ def runTestServer() {
       // runtests_server.py writes to this directory.  Make sure it's clean
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
+
+      tests = sh(script: "cat ../files_to_test.txt", returnStdout: true);
+      if (tests.isAllWhitespace()) {
+         echo("No Unit Tests to run!")
+         skipTestAnalysis = true
+         return
+      }
 
       // START THE SERVER!  Note this blocks.  It will auto-exit when
       // it's done serving all the tests.
@@ -529,6 +538,13 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
             runTests();
          }
       } finally {
+         // If we determined there were no tests to run, we should skip 
+         // analysis since it fails if there are no test results.
+         if (skipTestAnalysis) {
+            echo("Skipping Analysis - No tests run")
+            return
+         }
+
          // We want to analyze results even if -- especially if --
          // there were failures; hence we're in the `finally`.
          stage("Analyzing results") {


jobs/e2e-test.groovy
```diff
@@ -302,6 +305,19 @@ def runTestServer() {
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
 
+      tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
```
csilvers, line 308: I think you'll find it easier to use exec.outputOf() here.  Likewise below.

jobs/e2e-test.groovy
```diff
@@ -302,6 +305,19 @@ def runTestServer() {
...
+      // The runtests_server.py script with -n outputs all tests to run of 
+      // various types. We only need to worry about smoke tests, which are 
+      // also the last section, so grab everything after the header. If it's 
+      // empty, there are no tests to run.
+      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 13).trim()
+      echo("e2eTests: ${e2eTests}")
```
csilvers, line 314: Did you mean to keep this in?

jobs/e2e-test.groovy
```diff
@@ -302,6 +305,19 @@ def runTestServer() {
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
 
+      tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
+      // The runtests_server.py script with -n outputs all tests to run of 
+      // various types. We only need to worry about smoke tests, which are 
+      // also the last section, so grab everything after the header. If it's 
+      // empty, there are no tests to run.
```
csilvers, line 312: This is fragile! -- the output of `-n` isn't meant to be machine-readable.  I'd at least put a pointer in runtests_server.py pointing to this, if someone wants to change the output of the script.  And maybe make sure runtests_server promises that the smoketests are last.

More robust would be to do a something like this:
   testing/runtests_server.py -n <args> . | sed -n /^SMOKE TESTS:/,/:/p | grep -v $ | xargs

This will get everything between the "SMOKE TESTS" section and next section (or end of output), and trim whitespace.  (You could also keep the trimming in groovy, of course.) 
 Then the isAllWhitespace() check should work really reliably; you'd just need to comment in runtests_server.py not to change the labels of the test types.

jobs/webapp-test.groovy
```diff
@@ -278,6 +280,13 @@ def runTestServer() {
       // before it does so, so it doesn't read "old" data.
       sh("rm -rf genfiles/test-reports");
 
+      tests = sh(script: "cat ../files_to_test.txt", returnStdout: true);
```
csilvers, line 283: This version is definitely simpler!

yogieric (approved):
LGTM

csilvers (changes requested):
Thanks for doing this!  I like the solution you came up with.  "Request changes" because I think it's not difficult to make it less brittle.

I am a bit worried that `runtests_server.py -n` is not fast, so this slows down e2e tests in the common case.  (A major goal of the runtests-server refactor was to get rid of that test-reading overhead.)  Since e2e tests are on the critical path that's not the awesome.  But fixing this in a more principled way is annoying.

I think the way to do it would be to add a `/peek/<language>` endpoint to runtests_server.py that is the same as `/tests/<language>` but a) only ever serves up a single test, and b) does not "consume" the test for future /peek or /tests calls.  Then you can start the runtests-server, do a peek for smoke-tests, and if it returns no-more-tests, immediately do a /begin + /end to cause the server to quit, and then set `skipTestAnalysis` like you do now.

jobs/e2e-test.groovy
```diff
@@ -302,6 +305,19 @@ def runTestServer() {
...
+      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 13).trim()
+      echo("e2eTests: ${e2eTests}")
+      if (e2eTests.isAllWhitespace()) {
+         echo("No E2E Tests to run!")
+         skipTestAnalysis = true
+         return
```
csilvers, line 318: I think the run will be more effective if you do `throw new TestsAreDone()` here?  Likewise below.  It's unclear to me if this is still causing all the clients to start up or not.  Honestly, I am surprised the code you have is working as-is.  What is stopping the clients from starting up, trying to talk to the server, and being upset they can't?

I think at least you need to be doing this before you set TEST_SERVER_URL, not after.

It's hard for me to tell what testing you actually did.  It would be great if your test plan were a recipe that other people could follow to confirm that things work correctly.  Something like:
```
I started a jenkins e2e job with the old code and set TEST_TO_RUN to <something> and TESTS_TO_SKIP to <same thing>.  I ran the job and saw it errored out.  Then I clicked on "replay" and replaced the main e2e code with the code in this commit, and ran it again, and this time it didn't error out.
```
It would be good to verify that both runs take about the same amount of time, as well!  Or that the new code is faster. :-)

jobs/e2e-test.groovy
dbraley: I suppose that's possible. The `sh` function is pretty simple to use and comes with jenkins so has lots of great documentation and examples, whereas `exec.outputOf` is our own code with no testing or documentation that I can find. It does essentially what is done here as a convenience, but it looks to require it's arguments be provided as an array, though I never remember exactly how groovy switches between arg lists and arrays so perhaps if it's as trivial as `exec.outoutOf("testing/runtests_server.py -n ${runSmokeTestsArgs} .")` then I suppose it would be easier.

jobs/e2e-test.groovy
dbraley: I did, as it could definitely make the debugging easier, but the logs are darn noisy already so if we want to eliminate it we certainly could.

jobs/e2e-test.groovy
dbraley: > This is fragile! -- the output of `-n` isn't meant to be machine-readable. I'd at least put a pointer in runtests_server.py pointing to this, if someone wants to change the output of the script. And maybe make sure runtests_server promises that the smoketests are last.

I agree! A json output would certainly be more usable! I can write up a ticket to capture that work, but it's not in this repo so would really be better asynchronously done to this work.

> More robust would be to do a something like this:
> testing/runtests_server.py -n . | sed -n /^SMOKE TESTS:/,/:/p | grep -v $ | xargs
> 
> This will get everything between the "SMOKE TESTS" section and next section (or end of output), and trim whitespace. (You could also keep the trimming in groovy, of course.)
> Then the isAllWhitespace() check should work really reliably; you'd just need to comment in runtests_server.py not to change the labels of the test types.

I think we could get to that same level of robustness just by adding a further substring method that trims off everything after an empty newline (if there is one), and doing it in the groovy would probably be more readable than sed (at least to me, I suppose I can't speak to other maintainers). Also it appears there's a problem using pipes in the shell commands, I tried a similar solution to what you proposed here first and it consistently failed the whole job.

jobs/e2e-test.groovy
dbraley: > I think the run will be more effective if you do `throw new TestsAreDone()` here? Likewise below. It's unclear to me if this is still causing all the clients to start up or not. Honestly, I am surprised the code you have is working as-is. What is stopping the clients from starting up, trying to talk to the server, and being upset they can't?

The clients do fail (but don't fail the job), and I was trying to find a way to make them quit more quickly, so that's great if that works that simply. I was unaware of that. I'm moderately concerned it will be difficult to achieve the desired path without a lot of testing given the potential threading and race-condition implications, but it's probably worth a quick attempt.


> I think at least you need to be doing this before you set TEST_SERVER_URL, not after.

Oh, because setting that unblocks the workers. Good point, that's an. easy fix. Sorry I didn't see that comment. We could probably similarly exit the test workers before starting them if the `skipTestAnalysis` is false (though probably want to rename it rather than overload it like that).


> It's hard for me to tell what testing you actually did. It would be great if your test plan were a recipe that other people could follow to confirm that things work correctly. Something like:
> 
> > I started a jenkins e2e job with the old code and set TEST_TO_RUN to `<something>` and TESTS_TO_SKIP to `<same thing>`.  I ran the job and saw it errored out.  Then I clicked on "replay" and replaced the main e2e code with the code in this commit, and ran it again, and this time it didn't error out.

It's not clear to me how what you're asking for is different than the testing plan I've already provided, which includes both links to runs where this was tested, as well as instructions for how to replicate each scenario. I suppose I didn't call out the fact that the first run of each do-it-yourself option will result in a failure?


> It would be good to verify that both runs take about the same amount of time, as well! Or that the new code is faster. :-)

That's reasonable, though I will point out as we have found many times before that benchmarking jenkins on a job-by-job basis is unreliable, since so much of the processes have to wait on external factors. Q&D analysis (based on clicking the available passed jobs in jenkins):

Unit test take 10-35 minutes to run when they pass, with an average that looks to be in the 14-17 minute range. The test case where I ran the full suite of tests above with the code here [webapp-test 113728](https://jenkins.khanacademy.org/job/deploy/job/webapp-test/113728/) ran in 13 minutes, so just a hair under the average - not surprising, since it was doing exactly the same work other than a very fast bit of computation. The case where there are no unit tests [webapp-test 113727](https://jenkins.khanacademy.org/job/deploy/job/webapp-test/113727/) ran in 1 minute 7 seconds, so quite a bit faster (it's previous failures were similarly fast, fwiw).

E2e testing takes between 4 and 20 minutes based on a similar quick analysis. The test run above which had one test case,  [e2e-test 31199](https://jenkins.khanacademy.org/job/deploy/job/e2e-test/31199/) ran in 4m32s, so on par with the fastest runs. In only had one test to run, so that may not be a perfect picture, I'll make a run with this code that doesn't skip anything and see how it performs. The case where there were zero e2e tests to run [e2e-test 31202](https://jenkins.khanacademy.org/job/deploy/job/e2e-test/31202/) ran in 1m01s, so quite quickly.

That last number gives me optimism that the extra call to `testing/runtests_server.py -n` isn't costing all. that much time, though it's certainly possible it works significantly faster when there are no tests to run...

dbraley:
> I am a bit worried that `runtests_server.py -n` is not fast, so this slows down e2e tests in the common case. (A major goal of the runtests-server refactor was to get rid of that test-reading overhead.) Since e2e tests are on the critical path that's not the awesome. But fixing this in a more principled way is annoying.

That is a reasonable concern, though as pointed out in the last comment I suspect it's not a big impact to overall test execution time. Based on a quick parsing of the two logs, It looks like the call takes around 6 seconds, which is probably not a significant slow down when compared to the greater size of the e2e tests suite. Though again, it's possible that time goes up when the full suite of tests is being run, so I'll get those number shortly too.

> I think the way to do it would be to add a `/peek/<language>` endpoint to runtests_server.py that is the same as `/tests/<language>` but a) only ever serves up a single test, and b) does not "consume" the test for future /peek or /tests calls. Then you can start the runtests-server, do a peek for smoke-tests, and if it returns no-more-tests, immediately do a /begin + /end to cause the server to quit, and then set `skipTestAnalysis` like you do now.

As with the improvements to the runtests_server.py script to modify the -n output or provide a json output, such a solution requires making some significant changes in webapp first. I'm glad to write up follow-up tickets to do this work, but don't presently see a compelling reason to do that work in this task.

# Planned additional work
[ ] Get times analysis for a full e2e run to assure it falls without the expected time range and that the extra processing doesn't particularly slow down the job.
[ ] Move the test count analysis above the server IP setting to hopefully prevent the clients from flailing in vain, which adds confusing log messages and potentially causes the testing jobs to go longer (thus hopefully fixing that).
[ ] See if there's a quickly available and more elegant way to shut down the test suite when there are no tests to run by using the `throw new TestsAreDone()` exception
[ ] Try replacing `sh` call with `exec.outputOf` for code readability. Timebox to 5 minutes.

Let me know if I missed anything or you wish to see other changes.

jobs/e2e-test.groovy
alanszlosek: @dbraley :+1: to string manipulation in a programming language being more readable (and discoverable, even) than shelling out to sed.

dbraley:
Ok, I ran a test with the full suite of e2e tests enabled. The slowdown was at MOST about 7.5 seconds, based on the timestamps. The run itself was actually faster (12 minutes vs 9m51s) compared to the same run without this code, though I suspect that's more the random other variables going on, as there's no reason I see that this code would make the test suite faster. But I think it's safe to say it's probably not significantly slower. When we take into account that we get about 1 dev-support ticket per month about this, which usually delays actual deployment by 10-30 minutes, it's almost certainly worth it, especially as usually the only thing waiting on this test phase to end is a human.

jobs/e2e-test.groovy
```diff
@@ -290,6 +293,28 @@ def runTestServer() {
...
 
+      // Determine if we actually need to run any tests at all
+      // tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
+      tests = exec.outputOf(["testing/runtests_server.py", "-n", runSmokeTestsArgs, "."].flatten());
+
+      // The runtests_server.py script with -n outputs all tests to run of 
+      // various types. We only need to worry about smoke tests, which are 
+      // also the last section, so grab everything after the header. If it's 
+      // empty, there are no tests to run.
+      e2eTests = tests.substring(tests.lastIndexOf("SMOKE TESTS:") + 12)
```
dbraley, line 304: It may be worth noting this offset changed to 12 from 13. The 12 is the correct length of `SMOKE TESTS:`, 13 was just stripping a newline for us. That behavior prevented the `\n\n` from matching (since the first `\n` was removed) if there were no e2e tests and only a single line before a new section.

dbraley:
@csilvers care to take another look?

dbraley:
Oh! Forgot to drop links to the test runs;
(e2e-test 31353)[https://jenkins.khanacademy.org/job/deploy/job/e2e-test/31353/console] - Tested new e2e code with 1 e2e test, passed in **9m14s**
(e2e-test )[https://jenkins.khanacademy.org/job/deploy/job/e2e-test/31366/console] - Tested new e2e code with 0 e2e tests enabled, passed in **1m30s** after correctly skipping analysis
(webapp-test 114290)[https://jenkins.khanacademy.org/job/deploy/job/webapp-test/114290/console] - Tested new webapp-test code with 0 unit tests enabled, passed in **54s** after correctly skipping analysis

jobs/e2e-test.groovy
csilvers: It would be something like:
   exec.outputOf(["testing/runtests_server.py", "-n"] + runSmokeTestArgs)
It is not only easier, it is safer, because you don't have to worry about escaping.  Honestly, `sh` was a bad design decision on jenkins' part.  As you say, figuring out what `exec` does is not difficult, and it matches how exec works in most every programming language.

jobs/e2e-test.groovy
csilvers: I would say that's true for most programming languages, but not for groovy. :-/  That said, it's up to you.

> I think we could get to that same level of robustness just by adding a further substring method that trims off everything after an empty newline (if there is one)

That works too!

jobs/e2e-test.groovy
csilvers: No, it takes a constant amount of time to run because it has to read the whole filesystem.

When it comes to timing, I think what you'll want to do is to go to the "pipeline steps" page and see how long individual pipeline steps take.  Like see how long the clients are running for before and after your change.  And then you can see how long the extra runtests_server call is taking directly.

csilvers: > It's not clear to me how what you're asking for is different than the testing plan I've already provided, which includes both links to runs where this was tested, as well as instructions for how to replicate each scenario. I suppose I didn't call out the fact that the first run of each do-it-yourself option will result in a failure?

I think the key is to write a test plan that can be followed by someone who doesn't really understand the system.  So it needs to be very precise: not just "run the job" but "visit https://jenkins.khanacademy.org/job/deploy/job/e2e-test" and click on the "build now" button, and put "<whatever>" in the "version" field and click "build".  (And I usually frame it not as commands you should run but commands I ran: "I visited https://... and clicked "build"", though practices here differ.)  I think you pointed to a generic "how to use jenkins" page as a substitute for that, but that's not as good a substitute, since that page covers a lot of things that *aren't* appropriate for testing this change.

As a specific example, I don't understand from your test plan how you ran your new code jenkins-job, rather than the existing code.

jobs/e2e-test.groovy
```diff
@@ -290,6 +293,28 @@ def runTestServer() {
          runSmokeTestsArgs += ["--skip-tests=${params.SKIP_TESTS}"];
       }
 
+      // Determine if we actually need to run any tests at all
+      // tests = sh(script: "testing/runtests_server.py -n ${exec.shellEscapeList(runSmokeTestsArgs)} . ", returnStdout: true);
+      tests = exec.outputOf(["testing/runtests_server.py", "-n", runSmokeTestsArgs, "."].flatten());
```
csilvers, line 298: Oh!  I see how you did that.  You don't need the `flatten()`, you can just do:
```
tests = exec.outputOf(["testing/runtests_server.py", "-n"] + runSmokeTestsArgs + ["."])
```

csilvers: OK, agreed that 7 seconds is not the end of the world. :-)  But I still think it would appropriate to add a TODO to refactor things somehow so we don't need this extra call to runtests_server.py.

Or maybe do what we do for the unit-tests, and actually use this list as input to the next runtests_server.py call at line 333!  But that can be a TODO as well, since I don't want to make you go through another round of testing.

jobs/e2e-test.groovy
csilvers: I think `"SMOKE_TESTS:".length()` would be even clearer!

csilvers (approved):
Thank you for being so careful with all the testing!  I know how annoying it is to wait for all the test-turnaround.  I was hoping things would be faster for you since you're testing the case that no e2e tests actually have to be run; 7 minutes in that case is still ridiculous. :-(  A problem for another day.

Anyway, this all looks good to me now!  A few places I think you can clean up the groovy a bit, but nothing requiring another go-around.

jobs/e2e-test.groovy
dbraley: Oh, I see! I provided the link to the wiki article that describes how to run the steps in the message but that might not have been obvious since it was just labelled 'here'. I can add a better description.

PR #11, opened 2021-09-16
Add slack thread to deploy znd

somewhatabstract:
## Summary:
I want to direct ZND messaging to a slack thread, so I'm updating the job to support that.
I already checked the commands that it calls and they all support `--slack-thread` already.

## Test plan:
I'll be running a ZND with a thread ID and checking that it reports where it should.


csilvers (approved):
Great!

PR #19, opened 2021-11-22
Try to give more informative messages when jenkins jobs fail.

csilvers:
## Summary:
At this point, the console logs are not very useful for figuring out
what's going on with a jenkins job, since so many things are happening
in parallel.  In practice, I find looking at the pipeline steps to be
the way to figure out what went wrong.

In this commit, I change the generic "job failed" message to point
people to the pipeline-steps page.  I also made it so we no longer
show the tail of the logs, since the heuristic "if there was an error
it's probably near the bottom" no longer applies very often.

Issue: https://khanacademy.slack.com/archives/C02NMB1R5/p1637011218262600

## Test plan:
`groovy vars/notify.groovy`

But mostly I'll deploy this and then run a webapp-test job and make
sure there are no groovy errors.

### DIFF
--- a/vars/notify.groovy
+++ b/vars/notify.groovy
@@ -83,34 +83,6 @@ def _statusText(status, includeNumConsecutiveFailures=true) {
 }
 
 
-// Returns the last 50 lines of the logfile.  However, we omit the
-// text of commands run by notify() (below) when it's easy to do so,
-// since they are run after the job proper has already finished, and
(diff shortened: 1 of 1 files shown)


kphilipkhan (approved):
While I hope I don't have any build errors I need to look at soon.  I'm looking forward to identifying the problem quicker!

PR #21, opened 2022-01-04
Upload the go-style gateway config to gcs when we upload the js-style one.

csilvers:
## Summary:
graphql-gateway-2 (aka go-go-graphql-gateway) uses a different style
of config file than graphql-gateway (aka apollo).  We need to upload
them both.

I protect this upload in an "if exists" check so things don't suddenly
start to break when we move to only having one go gateway again.

Issue: https://khanacademy.atlassian.net/browse/INFRA-7074

## Test plan:
groovy jobs/deploy-znd.groovy

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -460,6 +460,11 @@ def deployToGatewayConfig() {
        exec(["make", "-C", "services/graphql-gateway",
              "deploy-gateway-config",
              "DEPLOY_VERSION=${NEW_VERSION}"]);
+       if (fileExists("services/graphql-gateway-2/Makefile")) {
+          exec(["make", "-C", "services/graphql-gateway-2",
+                "deploy-gateway-config",
+                "DEPLOY_VERSION=${NEW_VERSION}"]);
+       }
    }
 }
 
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -288,6 +288,11 @@ def deployToGatewayConfig() {
        exec(["make", "-C", "services/graphql-gateway",
              "deploy-gateway-config",
              "DEPLOY_VERSION=${VERSION}"]);
+       if (fileExists("services/graphql-gateway-2/Makefile")) {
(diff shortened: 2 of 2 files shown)


jobs/build-webapp.groovy
```diff
@@ -460,6 +460,11 @@ def deployToGatewayConfig() {
        exec(["make", "-C", "services/graphql-gateway",
```
MiguelCastillo, line 460: Should we add a TODO here to remove when JS is no longer used?

jobs/build-webapp.groovy
```diff
@@ -460,6 +460,11 @@ def deployToGatewayConfig() {
...
              "deploy-gateway-config",
              "DEPLOY_VERSION=${NEW_VERSION}"]);
+       if (fileExists("services/graphql-gateway-2/Makefile")) {
```
MiguelCastillo, line 463: Is this `Makefile` check to be super extra cautious?  Or is there some situation i which we don't expect the make file to exist?

MiguelCastillo (approved):
Changes look fine to me.  Just wondering about whether or not we need the extra conditional to check if the make file exists.

jobs/build-webapp.groovy
csilvers: The expectation is when we get rid of the js we'll rename the go gateway to be services/graphql-gateway, and get rid of services/graphql-gateway-2.  That's why I have the `if` check as well.  (I talk about this a bit in the commit message.)

MiguelCastillo: > The expectation is when we get rid of the js we'll rename the go gateway to be services/graphql-gateway, and get rid of services/graphql-gateway-2. That's why I have the `if` check as well. (I talk about this a bit in the commit message.)

Awesome! Yup - might be worth adding a note around that expectation.

jobs/build-webapp.groovy
MiguelCastillo: errr - yeah I see you the details in the description had the information I needed.  Tho I think it's helpful to add the context in the code itself. :)

PR #23, opened 2022-01-31
Make some changes to try to get "build-current-sqlite" working again.

csilvers:
## Summary:
The main one is to wait longer for graphql-gateway to start up.  In my
local tests it can take over a minute.  I'm suspecting that is the
cause of these errors in the logs:

> 13:18:54  ERRO[0010] Post "http://localhost:8102/graphql/CreateUserModelsInOtherServices": dial tcp 127.0.0.1:8102: connect: connection refused

However I can't be sure because it's hard to tell what service is
outputting what logs.  I add some logic here to prepend every line of
log output with its service, to ease debugging.

I also found a typo involving redis-pid.

Issue: https://jenkins.khanacademy.org/job/misc/job/build-current-sqlite/1216/console

## Test plan:
Fingers crossed -- will try running this again and see how it goes!

### DIFF
--- a/build_current_sqlite.sh
+++ b/build_current_sqlite.sh
@@ -1,7 +1,7 @@
-#!/bin/sh -ex
+#!/bin/bash -ex
 
 # A script to build an up-to-date current.sqlite by using various tools to seed
-# our dev environment state.  
+# our dev environment state.
 
 # Specifically those tools are:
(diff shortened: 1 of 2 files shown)


jgervais (approved):
Great idea with the sed call!

PR #27, opened 2022-02-09
Clean old graphql-gateway files from gcs.

csilvers:
## Summary:
We forgot to do this, and now tens of thousands of obsolete graphql
schema files are cluttering up our gcs.  Let's clean the directory up
every week.

I don't actually know how often old schema files are used, so I'm
pretty conservative: I keep the last 2 months' worth of schema files.
To be extra safe, I also keep the last 50 (not including znds), so if
we ever get to a world where we're going for months without updating
the schema files, we'll still keep them around.

Issue: https://khanacademy.atlassian.net/browse/INFRA-6062

## Test plan:
I ran the commands manually and they seemed to delete good things.

### DIFF
--- a/weekly-maintenance.sh
+++ b/weekly-maintenance.sh
@@ -202,6 +202,28 @@ clean_ka_translations() {
     done
 }
 
+clean_graphql_gateway_schemas() {
+    # Files have the format:
+    #    csdl-YYMMDD-HHMM-######.json(.gz)
+    #    csdl-znd-YYMMDD-HHMM-######.json(.gz)
+    #    v2-YYMMDD-HHMM-######.json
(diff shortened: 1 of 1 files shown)


dnerdy (approved):
Nifty. It's an interesting strategy building up the match patterns this way. TIL about gsutil `-m` and `-I`. Btw, date `-d` isn't available on macOS (there is a `-d` option but it does something different), but it seems like that's probably okay?

csilvers:
> Nifty. It's an interesting strategy building up the match patterns this way. TIL about gsutil `-m` and `-I`. Btw, date `-d` isn't available on macOS (there is a `-d` option but it does something different), but it seems like that's probably okay?

Yeah, jenkins runs on linux.

PR #31, opened 2022-02-28
Ensure that rollback code is of the right version

dat-boris:
## Summary:
During the rollback in our KRD [1] we noticed that we deploy a wrong
version of the dispatch.yaml. We can visibly see the traffic goes
to the new version of the graphql-gateway.

This is due to the local version of the code would still be on the
latest version, and some operations such as updating `dispatch.yaml`
will pick up the config from local code.

This diff ensure that we use
[1] https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/9990/

Issue: https://khanacademy.slack.com/archives/C8XGW76FQ/p1646079377461529

Test plan:

1. Merge this which should deploy this to jenkins
2. Create a fake deploy that have a harmless route redirection (e.g.
   redirecting a https://khan-academy.appspot.com/trace  to service-testbed)
3. Deploy, set-default and immediately force a rollback.
4. Validate that the dispatch route is not working after the rollback

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -802,7 +802,7 @@ def finishWithFailure(why) {
                // During the rollback, we need the local version of code to be
                // of the good version for some of the rollback operation
                // e.g. update dispatch.yaml
-               exec(["git", "checkout", "tag/${ROLLBACK_TO}"]);
+               exec(["git", "checkout", "tags/${ROLLBACK_TO}"]);
 
                // We use --bad-dict instead of --bad here because
                // there's no git tag yet for the bad version:
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -259,33 +259,33 @@ def _alert(def slackArgs, def interpolationArgs) {
    }
 }
 
-// This method filters a common 'DEPLOYER_USERNAME' into a series of comma 
-// seperated slack user id's for the purpose of passing into alert.py's 
+// This method filters a common 'DEPLOYER_USERNAME' into a series of comma
+// seperated slack user id's for the purpose of passing into alert.py's
 // '--slack' argument. For Example:
 // DEPLOYER_USERNAME: <@UMZGEUH09> (cc <@UN5UC0EM6>)
 // becomes: UMZGEUH09,UN5UC0EM6,
 // TODO(dbraley): Add parsing of `@` usernames and `#` channels if needed
 @NonCPS // for pattern & matcher
 def _userIdsFrom(def deployUsernameBlob) {
-   // Regex to specifically grab the ids, which should start with U and be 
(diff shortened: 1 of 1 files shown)


dat-boris:
Draft PR - Leaving this for discussion with @csilvers - I think that's what we need, but there might be some more complications since I don't know a lot about our current jenkins....

* do we need to update deps manually if we checkout new version?  
* From testing looks like the test tag `gae-220228-0839-edf7b9102b43` is available, but it is guaranteed?  Is the testing sound?
* how do we test?  I started with a test plan but not sure if it is sound.

jobs/deploy-webapp.groovy
```diff
@@ -799,6 +799,11 @@ def finishWithFailure(why) {
                  gitTag: GIT_TAG]);
          withSecrets() {
             dir("webapp") {
+               // During the rollback, we need the local version of code to be
+               // of the good version for some of the rollback operation
+               // e.g. update dispatch.yaml
+               exec(["git", "checkout", "tag/${ROLLBACK_TO}"]);
```
csilvers, line 805:
```suggestion
               exec(["git", "checkout", "tags/${ROLLBACK_TO}"]);
```

csilvers (approved):
Great, thank you!  One (very important) typo-fix but then I think it's good to go.

We should have some way of testing it.  Maybe do a trivial deploy and set-default and then abort and make sure the rollback works.

jobs/deploy-webapp.groovy
dat-boris: Ah thanks for spotting this!

dat-boris:
Thanks for the review!

> We should have some way of testing it. Maybe do a trivial deploy and set-default and then abort and make sure the rollback works.

Agree! I'm going to try this test plan: 

1.  Merge this which should deploy this to jenkins
2. Create a fake deploy that have a harmless route redirection (e.g.  redirecting a https://khan-academy.appspot.com/trace to service-testbed)
3. Deploy, set-default and immediately force a rollback. 
4. Validate that the dispatch route is not working after the rollback

Going to land this and run with the test plan.  Please do let me know if there are any follow up comments I can address after the landing.

csilvers:
By "route redirection" do you mean entry in dispatch.yaml?  Because you're not allowed to have appspot.com urls in dispatch.yaml.  So you'll need to find some other redirection to do.  I think it would be safe to add this one:
```
  - url: "*/graphql/hi-i-am-testing-dispatch-yaml*"
    module: service-testbed
```

dat-boris:
Thanks for clarifying!

> By "route redirection" do you mean entry in dispatch.yaml?

Ah yes indeed I mean `dispatch.yaml`!  I was planning to try with this audit https://github.com/Khan/webapp/pull/3909 with a test rollback.

dbraley (commented):
Late, but LGTM

PR #33, opened 2022-03-17
[🔥AUDIT🔥] Upload the new supergraphql gateway config to gcs when we upload the old js one.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
queryplanner (the replacement for graphql-gateway) uses a different
style of config file -- the new supergraph format rather than the
csdl.  Until the replacement is copmlete, we need to upload them both.

I protect this upload in an "if exists" check so things don't break
for znd's or other old branches.

This is the analogue of #21, when we were using the go gateway configs.

Issue: https://khanacademy.atlassian.net/browse/INFRA-7637

## Test plan:
Fingers crossed


dbraley (commented):
LGTM

PR #35, opened 2022-03-21
[🔥AUDIT🔥] Run prime-query-plan-cache via the shell symlink, not directly.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
I'd lke ot move prime-query-plan-cache from graphql-gateway-2 to
graphql-gateway.  We shouldn't need to change our jenkins jobs when we
do that!  We already have a symlink in tools/ that we can/should use
instead.

While in the area, I cleaned up a few other graphql-gateway references.

Issue: https://khanacademy.atlassian.net/browse/INFRA-7739

## Test plan:


dbraley (commented):
👍🏻

PR #36, opened 2022-03-23
Get rid of the sync-captions jenkins job.

csilvers:
## Summary:
It hasn't been run in over a year, so presumably whatever it's doing
is now done some other way (or not at all).

Issue: https://khanacademy.atlassian.net/browse/INFRA-7722

## Test plan:
None


dnerdy (commented):
Alrighty!

csilvers:
@dnerdy do you mind approving?

dnerdy (approved):
> @dnerdy do you mind approving?

Sure. If the job isn't used, it isn't used.

PR #39, opened 2022-03-24
[🔥AUDIT🔥] Remove graphql-gateway-2 from the list of required services.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Nothing in master users graphql-gateway-2 anymore, so we don't need to
start it up in this script.  Which is good, because soon that service
will be going away entirely!

Issue: https://khanacademy.atlassian.net/browse/INFRA-7739

## Test plan:
Fingers crossed -- we'll make sure build-current-sqlite succeeds tomorrow.


dbraley (commented):
👍🏻

PR #42, opened 2022-03-31
[🔥AUDIT🔥] Try to fix build_current_sqlite by being smarter about waiting for services.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
It's failed the last few days, seemingly because it wasn't waiting
long enough for the queryplanner to start up (it can take a while).
This change uses polling instead of just an arbitrary sleep, so
hopefully it will work better.

Issue: https://jenkins.khanacademy.org/job/misc/job/build-current-sqlite/1287

## Test plan:
Fingers crossed, will try running the job in prod.

### DIFF
--- a/build_current_sqlite.sh
+++ b/build_current_sqlite.sh
@@ -14,18 +14,28 @@
 #
 # For this script to work, secrets.py must be on the PYTHONPATH.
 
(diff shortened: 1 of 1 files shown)


dnerdy (commented):
👍

PR #44, opened 2022-04-18
When we run on jenkins-master, print that fact.

csilvers:
## Summary:
`onWorker` says what worker machine a job is running on, but
`onMaster` didn't say it was running on master.  This PR fixes that.

Issue: https://khanacademy.slack.com/archives/C8Y4Q1E0J/p1650297296717879

## Test plan:
groovy vars/onMaster.groovy


dbraley (approved):
Thanks Craig, LGTM

PR #46, opened 2022-05-25
Build new e2e-test-cypress job

rgagarinr:
## Summary:
This job runs lambdatest-cypress-cli in the webapp environment
depending on how parameters are specified.

Issue: https://khanacademy.atlassian.net/browse/FEI-4473

## Test plan:
run with Jenkins Replay


jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+
+def _setupWebapp() {
+   GIT_SHA1 = kaGit.resolveCommitish("git@github.com:Khan/webapp",
+                                     params.GIT_REVISION);
+   
+   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_SHA1);
+
+   dir("webapp") {
+      sh("make clean_pyc");
+
```
csilvers, line 87: I don't think you need this?  You don't run any python, do you?

jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+                             "e2e-test # (${params.URL}: ${params.REVISION_DESCRIPTION}) @${params.DEPLOYER_USERNAME}",
+                             "-p",
+                             "${params.NUM_WORKER_MACHINES}"
+    ];
+    
+    try {
+        dir('webapp/services/static') {
+            
+            sh("git checkout feature/cypress");
+            sh("yarn install")
```
csilvers, line 115: It's not good practice to pollute the webapp workspace like this.  In particular, I think this will fail if you happen to run on a worker machine that already has feature/cypress set up.  You should be using `safeSyncToOrigin` instead, like with webapp.  And do that in workspace-root, so cypress is a sibling dir to webapp instead of embedded inside it.

(And probably you should take a CYPRESS_GIT_REVISION so people can run a non-master version of cypress if they want.)

I don't know how `yarn lambdatest` will work in that situation -- does it assume that feature/cypress is checked out in a particular location?  If so you may need to mess with that so it can find cypress out-of-tree.

jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+
+            withEnv(["LT_USERNAME=${lt_username}",
+                     "LT_ACCESS_KEY=${lt_access_key}"]) {
+                sh("yarn lambdatest ${exec.shellEscapeList(runLambdaTestArgs)}")
```
csilvers, line 119: Just do:
```
exec(["yarn", "lambdatest"] + runLambdaTestArgs)
```
(though I'd probably just stick the `yarn` and `lambdaTest` in the runLambdaTestArgs list from the beginning, if it were me).

jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+            sh("yarn install")
+
+            withEnv(["LT_USERNAME=${lt_username}",
+                     "LT_ACCESS_KEY=${lt_access_key}"]) {
+                sh("yarn lambdatest ${exec.shellEscapeList(runLambdaTestArgs)}")     
+            }
+        }
+    } catch (e) {
+        throw new TestFailed("LambdaTest build failed!");
+    }
```
csilvers, line 124: I don't think you need to catch and rethrow here.  Just let any errors percolate up normally.

jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+}
+
+onWorker("ka-test-ec2", '6h') {
+
```
csilvers, line 128: (Nit: we don't typically have blank lines here.)

jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+   """How many worker machines to use. Max available is 20.""",
+   "20"
+
+).addStringParam(
+   "TEST_RETRIES",
+   """How many retry attempts to use. By default is 1.""",
+   "1"
+
+).addChoiceParam(
+   "TEST_TYPE",
```
csilvers, line 64: This is not used.  Did you want to take it out?  Or add a TODO somewhere to use it.

jobs/e2e-test-cypress.groovy
```diff
@@ -0,0 +1,138 @@
...
+  <li> <b>nightly</b>: run only night list of tests.</li>
+</ul>
+""",
+   ["all", "deploy", "nightly"]
+
+
+).apply()
+
+def _setupWebapp() {
+   GIT_SHA1 = kaGit.resolveCommitish("git@github.com:Khan/webapp",
```
csilvers, line 80: This isn't a global so you can just use `def`.  Though I think you can probably just call:
```
kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_REVISION);
```
?  You don't need to store the sha1 because you're not needing to use the same sha1 again elsewhere, like with the job you copied from.

csilvers (changes requested):
Awesome that you go this working!  I'm excited to see how it does.

"Request changes" so you're cloning the cypress repo under workspace/ and not inside webapp/.

rgagarinr:
Hi @csilvers, thanks for your detailed feedback, going to fix them all!

jobs/e2e-test-cypress.groovy
rgagarinr: Yeah, wasn't sure what this was for, thought some cleaning webapp stuff.

jobs/e2e-test-cypress.groovy
rgagarinr: Done, thanks!

jobs/e2e-test-cypress.groovy
rgagarinr: Removed!

jobs/e2e-test-cypress.groovy
```diff
@@ -18,14 +17,13 @@ import org.khanacademy.Setup;
...
 
 new Setup(steps
 
 ).addStringParam(
-   "GIT_REVISION",
-   """The name of a webapp branch to use when building.""",
-   "master"
+   "CYPRESS_GIT_REVISION",
+   """The name of a cypress branch to use when building.""",
+   "feature/cypress"
```
csilvers, line 26: Oh interesting.  Before you were checking out a separate repo but now you're not?  And `feature/cypress` is a branch on webapp?

jobs/e2e-test-cypress.groovy
```diff
@@ -73,66 +71,56 @@ Defaults to GIT_REVISION.""",
...
-                                     params.GIT_REVISION);
-   
-   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_SHA1);
-
-   dir("webapp") {
-      sh("make clean_pyc");
+   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", params.CYPRESS_GIT_REVISION);
 
+   dir("webapp/services/static") {
+      sh("yarn install")
```
csilvers, line 80: Better would be to do:
```
make -C webapp static_npm_deps
```
in case there's more to do for getting the deps set up right.  But this probably works too.

jobs/e2e-test-cypress.groovy
```diff
@@ -73,66 +71,56 @@ Defaults to GIT_REVISION.""",
...
-
-            withEnv(["LT_USERNAME=${lt_username}",
-                     "LT_ACCESS_KEY=${lt_access_key}"]) {
-                sh("yarn lambdatest ${exec.shellEscapeList(runLambdaTestArgs)}")     
-            }
-        }
-    } catch (e) {
-        throw new TestFailed("LambdaTest build failed!");
-    }
+   _setupWebapp();
```
csilvers, line 85: Normally we do this as a separate stage, so you'd have
```
stage("Sync webapp") {
   _setupWebapp();
}
stage("Run e2e tests") {
   ...
}
````
in your main section.  That just lets us get more accurate timing information on how long it takes to actually run the lambda test.

File: jobs/e2e-test-cypress.groovy [hunk 11]
```diff
@@ -73,66 +71,56 @@ Defaults to GIT_REVISION.""",
...
-                    when: ['FAILURE', 'UNSTABLE']]]) {
-
-        stage("Run e2e tests") {
-            runLamdaTest();
-        }
-    }
+      stage("Run e2e tests") {
+         runLamdaTest();
+      }
+   }
```
csilvers, line 123: (There are a few extra blank lines at the end of the file.)

csilvers (approved):
A few minor nits, but overall this looks great!  Assuming I'm correct you don't need to check out a second repo after all.

jobs/e2e-test-cypress.groovy
rgagarinr: btw, it is working well..

jobs/e2e-test-cypress.groovy
rgagarinr: Basically, we just need to install `lambdatest-cli` with dependencies inside lambdaconfig file, so we should be good.

jobs/e2e-test-cypress.groovy
csilvers: Yes, in that case what you have is exactly right.

PR #47, opened 2022-06-07
Notify Jenkins build failures

rgagarinr:
## Summary:
With this pr we’re trying to use lambda api’s to get build&number status after lambda server work is done (by using sync flag) then notify deployer if the build is failed. At the same time, we don’t want to send test logs to dev-support-log channel during integration.

Issue: FEI-4478

## Test plan:
Run the script with replay on both main and notify scripts

### DIFF
--- a/jobs/e2e-test-cypress.groovy
+++ b/jobs/e2e-test-cypress.groovy
@@ -7,6 +7,8 @@
 // This job runs lambdatest-cypress-cli in the webapp environment 
 // depending on how parameters are specified.
 
+// The Jenkins "interrupt" exception: for failFast and user interrupt
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
 
 @Library("kautils")
 // Classes we use, under jenkins-jobs/src/.
@@ -16,7 +18,6 @@ import org.khanacademy.Setup;
 //import vars.notify
 //import vars.withTimeout
 //import vars.withSecrets
-
 
 new Setup(steps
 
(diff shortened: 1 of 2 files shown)


csilvers:
I don't understand the motivation for this change.  The `notify` rule already notifies the user when the build fails.  It seems like this PR is mostly recreating that logic.  What is going on that you can't just use the existing `notify`?

it seems like the BUILD_NAME stuff is separate from the notification stuff.  I think it's probably going to be easiest to separate them into two separate PRs.

rgagarinr:
Spent too much time on API workaround...  :(

jobs/e2e-test-cypress.groovy
```diff
@@ -92,28 +96,42 @@ def runLamdaTest() {
...
+   // TODO(ruslan): Remove @ in extraText before merging with current e2e-test pipeline, 
+   // we don't want to disturb real users.  
    notify([slack: [channel: params.SLACK_CHANNEL,
-                  when: ['FAILURE', 'UNSTABLE']]]) {
+                   sender: 'Testing Turtle',
+                   emoji: ':turtle:',
+                   extraText : 'Hey '+'@' + 
+                   params.DEPLOYER_USERNAME + 
+                   " BUILD: " + BUILD_NAME + 
+                   ') FAILED (<https://automation.lambdatest.com/logs/|Open>)',
```
csilvers, line 133: Doesn't notify.groovy already say that the build failed?  I'm worried this is just going to repeat information that's already there, and the only new thing will be the lambdatest link.  But we can always tweak this text later.

It would be awesome, btw, the if the link could go directly to the logs for this run, rather than the generic "logs" page.  Is that posible?

jobs/e2e-test-cypress.groovy
```diff
@@ -7,6 +7,8 @@
 // This job runs lambdatest-cypress-cli in the webapp environment 
 // depending on how parameters are specified.
 
+// The Jenkins "interrupt" exception: for failFast and user interrupt
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
```
csilvers, line 11: You can remove this; you don't use this symbol anymore.

csilvers (approved):
I like it, nice and simple!  Once you remove the import of FlowInterruptedException this is good to go, to me.

jobs/e2e-test-cypress.groovy
```diff
@@ -76,49 +77,107 @@ Defaults to GIT_REVISION.""",
...
+def notifyFailures() {
+   def lambdaURL="https://api.lambdatest.com/automation/api/v1/builds"
+      
+   // Calling to LambdaTest for build name and id values,
+   // it will return 10 last build records.
+   // Retry 2 times if connection refused?
+   def response = sh(
+      script: "curl -s --retry-connrefused --retry 2 --retry-delay 5 " +
+              "${lambdaURL} -u ${LT_USERNAME}:${LT_ACCESS_KEY}", 
+      returnStdout: true).trim();
```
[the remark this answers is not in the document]
rgagarinr: oh, I see thanks!

jobs/e2e-test-cypress.groovy
rgagarinr: Yeah, `notify.groovy` part is responsible for the first chunk like: 
```
deploy/e2e-test-cypress failed #59 ([Open](https://jenkins.khanacademy.org/job/deploy/job/e2e-test-cypress/59/))
deploy/e2e-test-cypress failed: See https://jenkins.khanacademy.org/job/deploy/job/e2e-test-cypress/59/flowGraphTable/ for what failed (look for the red circle(s)).
```
And with the rest block, we might want to ping the user directly. The only similar custom `_alert` function I found in `deply-webapp.groovy` 

Concerning putting the link directly to the logs for this run, yeah it was working on the previous PR version:
```
BUILD_ID = sh(
      script: "echo '$response' | jq -r '.data | map(select(.name ==" +
              " \"${BUILD_NAME}\"))[0].build_id'", 
      returnStdout: true).trim();
```  
Let me try a few tweaks on how to implement this API to the current version. The tricky part is we have to first execute `lambda-cli` and after do a call to get freshly generated `build_id` that we can use in the link.

rgagarinr:
> I like it, nice and simple! Once you remove the import of FlowInterruptedException this is good to go, to me.

Thanks Craig really appreciate your help!

jobs/e2e-test-cypress.groovy
rgagarinr: Ok, I was able to get that BUILD_ID with `60s` delay parallel job in `run-notify`: 
```
parallel([
         "run-lambda": { exec(runLambdaTestArgs); },
         "run-notify": { notifyFailures(); }, // wait for ~60 sec to have generated lambdatest build_id 
])
```
But the problem is how to pass BUILD_ID in the upstream `notify.groovy` wrapper from the stage... obviously, it returns null?

Not sure, but looks like we have to create some [custom notify function](https://github.com/Khan/jenkins-jobs/pull/47/commits/491a18830186f0331ff5d545baaa7c6f2a2be926#diff-34a902da56226c29ae9f839e49290d303ad18000e36ea6096698afcbbf48be63R145) to have this lambda logs link?
@csilvers WDYT?

jobs/e2e-test-cypress.groovy
```diff
@@ -92,28 +93,42 @@ def runLamdaTest() {
...
+   // TODO(ruslan): Remove @ in extraText before merging with current e2e-test pipeline, 
+   // we don't want to disturb real users.  
    notify([slack: [channel: params.SLACK_CHANNEL,
-                  when: ['FAILURE', 'UNSTABLE']]]) {
+                   sender: 'Testing Turtle',
+                   emoji: ':turtle:',
+                   extraText : 'Hey '+'@' + 
+                   params.DEPLOYER_USERNAME + 
+                   " BUILD: " + BUILD_NAME + 
+                   ') FAILED (<https://automation.lambdatest.com/logs/|Open>)',
```
csilvers, line 130: This will be a lot more readable if you use string interpolation:
```
"Hey @${params.DEPLOYER_USERNAME} build ${BUILD_NAME} FAILED (<https://automation.lambdatest.com/logs/|Open>)"
```

jobs/e2e-test-cypress.groovy
csilvers: I'd say: just put in a TODO for now.  We can iterate on it over time, but let's get live what you have now.

rgagarinr: Sounds good!

PR #49, opened 2022-06-08
Add fake-gcs to the list of services started to build current.sqlite.

csilvers:
## Summary:
We can't run `make serve` in build_current_sqlite.sh because, um, it
didn't work for some reason.  I think INFRA-7279 has more details.  In
any case, this means we need to list the services to start manually.

It's easy to forget to do this, and indeed we did forget when we added
fake-gcs as a service-dependency in #5869.  This change adds it there.

Note that we will have to get mkcert working no the jenkins machines
as well.  That will be a separate diff (in a separate repo).

Issue:  https://khanacademy.slack.com/archives/C01120CNCS0/p1654719166232869

## Test plan:
Fingers crossed.

### DIFF
--- a/build_current_sqlite.sh
+++ b/build_current_sqlite.sh
@@ -28,9 +28,10 @@ function poll_for_service() {
 }
 
 function wait_for_service() {
-    # TODO(csilvers): poll something instead of just waiting for the first two.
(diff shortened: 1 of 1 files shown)


dnerdy (approved):
Aha! Thanks for tracking this down.

PR #53, opened 2022-06-23
Configure lambdatest run arguments

rgagarinr:
## Summary:
Issue: FEI-4628

## Test plan:
Run job with Replay

### DIFF
--- a/jobs/e2e-test-cypress.groovy
+++ b/jobs/e2e-test-cypress.groovy
@@ -99,16 +99,14 @@ def runLamdaTest() {
    // TODO(ruslan): Implement TEST_TYPE param to use decorator or flag 
    // in cypress script files. 
    // TODO(ruslan): Use build tags --bt with prod/znd states.
+   // TODO(ruslan): Add retries flag.
    def runLambdaTestArgs = ["yarn",
                             "lambdatest",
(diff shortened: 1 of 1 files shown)


jobs/e2e-test-cypress.groovy
```diff
@@ -99,16 +99,14 @@ def runLamdaTest() {
    // TODO(ruslan): Implement TEST_TYPE param to use decorator or flag 
    // in cypress script files. 
    // TODO(ruslan): Use build tags --bt with prod/znd states.
+   // TODO(ruslan): Add retries flag.
```
jandrade, line 102: Oh so that breaks when using single quotes?

jandrade (approved):
Looks great!

jobs/e2e-test-cypress.groovy
rgagarinr: yeah, I'm not sure why it's failing when [separating](https://docs.cypress.io/guides/guides/command-line#cypress-run-config-lt-config-gt) multiple values with commas. I'll check with the lambda team.

```
Cypress encountered an error while parsing the argument: --config

You passed: baseUrl="https://prod-xxyyy.khanacademy.org",

The error was: Cannot read properties of undefined (reading 'split')
```

rgagarinr:
Thanks for helping @jandrade

PR #54, opened 2022-07-06
Run LT/Cypress e2e tests sequentially.

jandrade:
## Summary:

There's an issue with the existing logic where we try to parallelize
Python/Selenium and Cypress/LambdaTest e2e tests. For some unknown
reason, LT is not being able to start some tests causing the entire test
suite to fail, and by using `parallel`, we're also seeing the main e2e
job having to wait until LT resolves (it's currently resolving after 1
hour of timeout).

This PR fixes that issue by running the LT/Cypress tests sequentially
only after the main e2e tests are completed.

Issue: XXX-XXXX

## Test plan:

Verify that the e2e job iss reporting times as before.

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -598,10 +598,7 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
 
       try {
          stage("Running smoketests") {
-            parallel([
-               "run-tests": { runTests(); },
-               "test-lambdacli": { runLambda(); },
-            ]);
+            runTests();
          }
       } finally {
          // If we determined there were no tests to run, we should skip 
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -616,6 +613,15 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
          stage("Analyzing results") {
             analyzeResults();
```
jeresig, line 614: Ah ha, ok! So I think the critical thing here is that the results will get reported from run-tests, which will unblock the deploys, and then the cypress tests will run after, separately.

jeresig (approved):
Great, thanks for fixing this!

jobs/e2e-test.groovy
dbraley: I'm not sure if that's quite correct... It's possible i'm wrong, but I think we need to exit the `notify` block, i think it's the `buildmaster` bit of that that kicks off the message to users to finish up?

jandrade: hmm yeah that makes sense. Maybe @csilvers also have a good idea about that?

jeresig (approved):
Great! Let's hope this works!

dbraley (commented):
👍🏻

PR #58, opened 2022-07-22
FEI-4674: Merge in 'master' when deploying ZNDs

kevinbarabash:
## Summary:
The regular deploy process merges 'master' into the deploy branch before deploying.  We want to do the same when deploying ZNDs.  This will avoid issues that can occur if a ZND is being deployed with out-of-date deploy scripts.

Issue: FEI-4674

Questions:
- Do we want to change GIT_REVISION to GIT_REVISIONS and allow people to deploy a ZND using multiple branches?

## Test plan:
- visit a page like https://jenkins.khanacademy.org/job/deploy/job/deploy-znd/5254/
- click on replay on the left pane. Replace the contents of the "main script" window with your new deploy-znd.groovy code
- Replce the contents of the "kaGit" window with your new kaGit.groovy code
- click "run" and see how the job does

```
...
17:02:51  'kevinb-fixes' resolves to bd2f7ca76da4b854c98f24bb340072c73be64a61
[Pipeline] sh
17:02:51  + git merge bd2f7ca76da4b854c98f24bb340072c73be64a61
17:02:51  Already up to date.
[Pipeline] sh

17:02:51  + git tag --points-at HEAD
[Pipeline] sh

17:02:52  + git rev-parse HEAD
[Pipeline] echo
17:02:52  Resolved master+kevinb-fixes --> 68711adc564fdb1a23685c4c93dacfd02ded95ef
...
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
```

- [x] repeat with a deploy-znd job that's been run for a branch that will have a merge conflict with `master`

https://jenkins.khanacademy.org/job/deploy/job/deploy-znd/5283 shows a deploy-znd failure due to merge conflict.  The error is reported in slack as a generic failure, but it's easy enough to track down the issue in the logs.  I'm going to leave it like this for now, but may circle back later to add a custom Slack message.  Here's what the logs looked like for the failure:
```
10:08:11  Failure message: Failed to merge 4b466c3815e7033565b72b1916ef3bcef8f472f2 into master: hudson.AbortException: script returned exit code 1
[Pipeline] ansiColor
[Pipeline] {
10:08:11  
[Pipeline] echo
10:08:11  ===== JOB FAILED =====
```

PR #61, opened 2022-07-25
Enforce ZND to use lowercase version name.

dat-boris:
## Summary:
As discovered during our ZND testing, if a ZND contains uppercase,
we will fail to lookup the static version in Fastly (see slack thread in issue).

Also from
https://github.com/Khan/jenkins-jobs/commit/d2dc0da5d3418682e222e3b9e2238d058bbb5864
it looks like previously we expect lowercase also.  So let's make the
version name lowercase.

Issue: https://khanacademy.slack.com/archives/C02NMB1R5/p1658776260277369

## Test plan:
- Create a new ZND job using uppercase ZND name https://jenkins.khanacademy.org/job/deploy/job/deploy-znd/5292/
- Select replay from above job and copy and paste the modified deploy-znd.groovy, and Run the job https://jenkins.khanacademy.org/job/deploy/job/deploy-znd/5293/.
- Confirm the new version https://prod-znd-220725-boris-up.khanacademy.org/_api/version is indeed working.

image

(note the different in the case of the ZND)

### DIFF
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -168,7 +168,7 @@ def determineVersion() {
    // VERSION parameter needs to be lowercased.
    // Otherwise Fastly will have issue looking up the static version as it expects
    // lowercase hostname.
-   VERSION = "znd-${date}-${_currentUser()}-${params.VERSION.toLowerCase()}";
+   VERSION = "znd-${date}-${_currentUser()}-${params.VERSION}".toLowerCase();
 
    // DNS has a limit of 63 bytes per hostname-component.  This
    // version can yield hostname-components like
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -165,7 +165,10 @@ def determineVersion() {
    }
 
    def date = new Date().format("yyMMdd");
-   VERSION = "znd-${date}-${_currentUser()}-${params.VERSION}";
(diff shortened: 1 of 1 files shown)


jobs/deploy-znd.groovy
```diff
@@ -165,7 +165,10 @@ def determineVersion() {
    }
 
    def date = new Date().format("yyMMdd");
-   VERSION = "znd-${date}-${_currentUser()}-${params.VERSION}";
+   // VERSION parameter needs to be lowercased.
+   // Otherwise Fastly will have issue looking up the static version as it expects
+   // lowercase hostname.
+   VERSION = "znd-${date}-${_currentUser()}-${params.VERSION.toLowerCase()}";
```
kevinbarabash, line 171: **suggestion (non-blocking):** maybe call `toLowerCase()` on the whole string afterwards.  I'm not sure if `_currentUser` is always lower case or not.

kevinbarabash (approved):
The changes seem reasonable to me.  Can you "replay" the ZND deploy that had an issue using a test plan similar to what was used for #58?  That way you can test it before landing this PR.

dat-boris:
Thanks for the tips @kevinbarabash !  TIL about the reply feature, superhelpful for the testing.

Will test with https://jenkins.khanacademy.org/job/deploy/job/deploy-znd/5243/replay/

yogieric (approved):
LGTM

PR #63, opened 2022-08-15
e2e-test-cypress Jenkins job is not running against current commit hash

rgagarinr:
## Summary:
We are trying to deploy a fix for a flaky test but LT still thinks that is an error
This is happening because we are using master to run the tests, so we are testing the “old” version of that test,
instead of the commit that includes the changes.

Issue: FEI-4724

## Test plan:
run replay

### DIFF
--- a/jobs/e2e-test-cypress.groovy
+++ b/jobs/e2e-test-cypress.groovy
@@ -22,7 +22,7 @@ new Setup(steps
 
 // Allow multiple LambdaTest runs to execute concurrently.
 ).addStringParam(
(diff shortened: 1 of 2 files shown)


jandrade (approved):
Looks great to me! We could probably introduce this too to the cypress backup job, but no need to do it as part of this PR.

Thanks for the quick fix! 🚢

rgagarinr:
> Looks great to me! We could probably introduce this too to the cypress backup job, but no need to do it as part of this PR.
> 
> Thanks for the quick fix! 🚢

Thanks Juan! 🤞

PR #66, opened 2022-09-02
[🔥AUDIT🔥] Always refer to the current revision by sha, not name.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
`GIT_REVISION` could be a symbolic name, like `master` or `csilvers`.
We convert it to a sha once at the beginning of webapp-test, and use
the sha everywhere in the script, to protect against changes in the
branch between the time the script starts and when it ends.  But we
missed one place that was still using GIT_REVISION.  Now it's fixed
too.

Issue: https://jenkins.khanacademy.org/job/deploy/job/webapp-test/150581/consoleText

## Test plan:
I ran a job successfully with these changes:
   https://jenkins.khanacademy.org/job/deploy/job/webapp-test/150583/console

### DIFF
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -243,10 +243,10 @@ def runTestServer() {
          // changed between BASE_REVISION and GIT_REVISION.  We ignore
          // files where only sync tags have changed; those can't affect
          // tests.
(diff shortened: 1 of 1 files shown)


MiguelCastillo:
good catch!

PR #69, opened 2022-09-29
Figure out what tests need to be run for deleted files, too.

csilvers:
## Summary:
We had an issue where there was a deploy that deleted one python file,
which made a second python file unused (since the deleted python file
was the only client).  But we didn't catch the problem because jenkins
doesn't call all_tests_for.py on deleted files.  On the face of it
that's reasonable: what tests could you possibly have on code that
doesn't exist?!  But there are some consistency tests that care, such
as unused_file_test.py.

This PR tells jenkins to look for that case, using a new
`--deleted-mode` flag on all_tests_for.py.

Depends on https://github.com/Khan/webapp/pull/8816.

Issue: https://khanacademy.slack.com/archives/C02NMB1R5/p1663856692389019

## Test plan:
Will run a webapp-test job using this new arg once https://github.com/Khan/webapp/pull/8816 is landed.

### DIFF
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -245,6 +245,8 @@ def runTestServer() {
          // tests.
          sh("deploy/trivial_diffs.py ${exec.shellEscape(params.BASE_REVISION)} ${exec.shellEscape(GIT_SHA1)} > ../trivial_diffs.txt");
          sh("git diff --name-only --diff-filter=ACMRTUB ${exec.shellEscape(params.BASE_REVISION)}...${exec.shellEscape(GIT_SHA1)} | fgrep -vx -f ../trivial_diffs.txt | testing/all_tests_for.py - > ../files_to_test.txt");
+         // Sometimes we need to run some extra tests for deletd files.
+         sh("git diff --name-only --diff-filter=D ${exec.shellEscape(params.BASE_REVISION)}...${exec.shellEscape(GIT_SHA1)} | fgrep -vx -f ../trivial_diffs.txt | testing/all_tests_for.py --deleted-mode - >> ../files_to_test.txt");
          // Note that unlike for tests, we consider deleted files for linting.
          sh("git diff --name-only --diff-filter=ACMRTUBD ${exec.shellEscape(params.BASE_REVISION)}...${exec.shellEscape(GIT_SHA1)} | testing/all_lint_for.py - > ../files_to_lint.txt");
       } else {


MiguelCastillo (approved):
Seems reasonable to me.

BTW I took the liberty to fix the links in your description to the other PRs.  But definitely feel free to revert if that's not what you wanted.

jobs/webapp-test.groovy
```diff
@@ -245,6 +245,8 @@ def runTestServer() {
          // tests.
          sh("deploy/trivial_diffs.py ${exec.shellEscape(params.BASE_REVISION)} ${exec.shellEscape(GIT_SHA1)} > ../trivial_diffs.txt");
          sh("git diff --name-only --diff-filter=ACMRTUB ${exec.shellEscape(params.BASE_REVISION)}...${exec.shellEscape(GIT_SHA1)} | fgrep -vx -f ../trivial_diffs.txt | testing/all_tests_for.py - > ../files_to_test.txt");
+         // Sometimes we need to run some extra tests for deletd files.
+         sh("git diff --name-only --diff-filter=D ${exec.shellEscape(params.BASE_REVISION)}...${exec.shellEscape(GIT_SHA1)} | fgrep -vx -f ../trivial_diffs.txt | testing/all_tests_for.py --deleted-mode - >> ../files_to_test.txt");
```
dbraley, line 249: Nit (or a question): Will the trivial_diffs.txt potentially contain deletions? if not, it seems unnecessary; if so, that would trim those out before running the command?

dbraley (approved):
I think this is good, so long as `trivial_diffs.txt` doesn't contain deleted files

jobs/webapp-test.groovy
csilvers: It's a little hard to reason about that.  In theory, trivial_diffs.txt could contain a deleted file if it *only* consisted of sync-tags, with nothing between them.  But obviously that would never happen in practice.  If it did, it would be safe to filter it out since a file that consists solely of sync tags can't depend on another file.

I'd like to keep the code the way it is for symmetry, if nothing else.  It may be that the trivial-diff filtering is unnecessary, but I feel this block of code is easier to reason about, as a whole, if we keep lines 247 and 249 as similar as possible.

csilvers:
> Seems reasonable to me.
> 
> BTW I took the liberty to fix the links in your description to the other PRs. But definitely feel free to revert if that's not what you wanted.

Oops!  Thanks for fixing the links.

jobs/webapp-test.groovy
dbraley: Cool, thanks for the clarification! Good by me!

yogieric (approved):

PR #73, opened 2022-11-30
Update decryption of webapp's secrets.py

aag:
This commit makes several changes to the decryption of webapp's secrets.py:
- Changes the password filename to the one in the latest aws-config scripts
- Changes the name of the encrypted file to the current one in the webapp repo
- Changes the openssl parameters to work with the current encrypted file in the webapp repo

Issue: https://khanacademy.atlassian.net/browse/TP-7594

### DIFF
--- a/vars/withSecrets.groovy
+++ b/vars/withSecrets.groovy
@@ -10,7 +10,7 @@ import groovy.transform.Field
 @Field _activeSecretsBlocks = 0;
 
 def _secretsPasswordPath() {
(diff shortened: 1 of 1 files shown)


csilvers (approved):
Looks good!  Glad you only have to change this in one place.  As I mentioned in the other PR, you'll want to wait until tomorrow to land this, so the workers have picked up the new image with secrets.py.aes.password in t.

yogieric (approved):

PR #74, opened 2022-12-02
FEI-4867: Only run Python/Selenium tests when the tests are retried

jandrade:
## Summary:

We are currently running Python/Selenium (Py/Sel) and Cypress/LambdaTest (Cy/LT) tests in
parallel. This happens everytime the `e2e-test` job is triggered.

While debugging some Cypress/LambdaTest test builds I noticed that the Cy/LT
tests are also triggered when we retry the Py/Sel tests in Slack (via
`buildmaster`).

This change attempts to run only Py/Sel tests when a set of Python e2e tests
fail, and prevent running Cy/LT builds that are not related to this retry.

Issue: FEI-4867

## Test plan:

In deployments, verify that the next time someone retries e2e tests, these don't
retry a LambdaTest build (or spin up a new `e2e-test-cypress` job).

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -603,10 +603,22 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
 
       try {
          stage("Running smoketests") {
-            parallel([
-               "run-tests": { runTests(); },
-               "test-lambdacli": { runLambda(); },
-            ]);
+            // In case we are retrying smoke tests, we should only run the
+            // Python/Selenium tests.  
+            // NOTE: `custom` is a special type that allow us to run a subset of
+            // tests. This happens when:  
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -603,10 +603,22 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
 
       try {
          stage("Running smoketests") {
-            parallel([
-               "run-tests": { runTests(); },
-               "test-lambdacli": { runLambda(); },
-            ]);
+            // In case we are retrying smoke tests, we should only run the
+            // Python/Selenium tests.
```
csilvers, line 607: Can you describe _why_ we don't want to run the cypress tests in this case?

csilvers (approved):
Looks good!

PR #81, opened 2023-01-24
Allow jenkins to update index/queue/pubsub.yaml outside deploy_to_gae.py.

csilvers:
## Summary:
Historically, when there have been changes to index.yaml, queue.yaml,
or pubsub.yaml, that's prompted a deploy of the default module (even
if the python code itself hasn't changed!), and deploy_to_gae.py would
update those yaml files in addition to updating the default module.

Now that python is going away, we want to separate those
responsibilities.  The plan is to create three new pseudo-services,
index.yaml, queue.yaml, and pubsub.yaml.  And should_deploy.py can say
we need to deploy to those services.  Jenkins notices this and uploads
the relevant yaml file to GAE.

This is the jenkins part of things: it notices when the new
pseudo-services are being deployed and uploads the yaml files.  Once
this is tested, I will change should_deploy.py to emit these
pseudo-services when appropriate, and change deploy_to_gae.py to no
longer try to upload to them.

Issue: INFRA-XXXX

## Test plan:
I ran `groovy jobs/build-webapp.groovy` with no new failures.  But my
main test will be to do a deploy after this is in, like so:
```
sun: queue csilvers (SERVICES=queue.yaml,pubsub.yaml,index.yaml)
```
and make sure it runs the upload commands with success.

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -437,7 +437,7 @@ def deployIndexYaml() {
 
 // This should be called from within a node().
 def deployQueueYaml() {
-   if (!("queue.yaml" in SERVICES)) {
+   if (!("queue_yaml" in SERVICES)) {
       return;
    }
    withSecrets() {    // TODO(csilvers): do we actually need Python secrets?
@@ -449,7 +449,7 @@ def deployQueueYaml() {
 
 // This should be called from within a node().
 def deployPubsubYaml() {
-   if (!("pubsub.yaml" in SERVICES)) {
+   if (!("pubsub_yaml" in SERVICES)) {
       return;
    }
    withSecrets() {    // TODO(csilvers): do we actually need Python secrets?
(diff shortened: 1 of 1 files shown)


jobs/build-webapp.groovy
```diff
@@ -67,12 +67,14 @@ options), or the special value "auto", which says to choose the services to
 deploy automatically based on what files have changed.  For example, you might
 specify "dynamic,static" to force a full deploy to GAE and GCS.</p>
 
-<p>Here are the services:</p>
+<p>Here are some services:</p>
 <ul>
   <li> <b>dynamic</b>: Upload dynamic (e.g. py) files to GAE. </li>
   <li> <b>static</b>: Upload static (e.g. js) files to GCS. </li>
   <li> <b>kotlin-routes</b>: webapp's services/kotlin-routes/. </li>
```
latteier, line 74: maybe remove this one?

jobs/build-webapp.groovy
```diff
@@ -552,8 +590,9 @@ def deployAndReport() {
          // in parallel, and hence must be handled sequentially. Those
          // deployments are bundled in deployToKotlinServicesAndDataflow().
          if (!(service in [
-               'dynamic', 'static', 'course-editing', 'kotlin-routes',
-               'dataflow-batch'])) {
+               'dynamic', 'static',
+               'course-editing', 'kotlin-routes', 'dataflow-batch',
```
latteier, line 594: maybe remove 'kotlin-routes' here too?

latteier (commented):
Thanks! I was wondering how this was going to happen with python not being deployed by default.

I don't have a lot of context here, but it looks reasonable to me.

MiguelCastillo (approved):
This makes sense to me!

PR #83, opened 2023-01-24
[🔥AUDIT🔥] Fix the comments to be more accurate.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
I claimed we update queue.yaml in build-webapp and not deploy-webapp,
but that's not really true: we do it in both places.  In the first we
add things, in the second we delete.  This updates the comments to be
clearer.

Issue: none

## Test plan:
none


MiguelCastillo (commented):
thanks!

dbraley (commented):
👍🏻

PR #85, opened 2023-02-07
[🔥AUDIT🔥] Point to the new location for some publish scripts.

csilvers:
🖍 _This is an audit!_ 🖍

I thought I had put in symlinks from the old location to the new
during this transition period, but I guess not.  We'll just update.

Issue: https://khanacademy.slack.com/archives/C49296Q7P/p1675116384824979

Test plan:
Will try deploying a new publish image


mpolyak (commented):
Thanks!

dbraley (commented):
Thanks!

PR #86, opened 2023-02-08
Add one default worker for priming Cypress e2e tests

jandrade:
## Summary:
We are currently priming 20 workers that are used by the Python/Selenium
e2e tests. We noticed that the Cypress e2e tests are taking ~2 minutes
before a worker becomes available so the job starts running and this is
due to the Python e2e tests taking up all the available workers.

This PR adds one default worker that will be used by the Cypress e2e
tests so that both `e2e-test` and `e2e-test-cypress` jobs can start
around the same time.

Note that Python and Cypress e2e tests are run in parallel, so we need
to prime+run then at the same time.

Issue: XXX-XXXX

## Test plan:

I've tested all the Jenkins jobs that are affected by this change using
the Replay feature. I checked that these changes didn't introduce any
regression:

1. Ran the `firstinqueue-priming` job first to prime the machines: https://jenkins.khanacademy.org/job/deploy/job/firstinqueue-priming/2761/console

2. Ran an `e2e-test `replay: https://jenkins.khanacademy.org/job/deploy/job/e2e-test/45423/console
3. This originated a new `e2e-test-cypress` job: https://jenkins.khanacademy.org/job/deploy/job/e2e-test-cypress/6912/console

I compared the `e2e-test` and the `e2e-test-cypress` jobs and noticed that the difference of start times when down to `~15 seconds`.

### Results
Current job (120 seconds) - PR changes (15 seconds) = `1m 45s gains`.  This means that the Cypress e2e job should finish around 1m 45s earlier than before.

### DIFF
--- a/jobs/e2e-test-cypress.groovy
+++ b/jobs/e2e-test-cypress.groovy
@@ -1,6 +1,6 @@
 // Single job for cypress e2e tests.
-// 
-// cypress e2e tests are the smoketests run in the webapp/feature/cypress repo, 
+//
+// cypress e2e tests are the smoketests run in the webapp/feature/cypress repo,
 // that hit a live website using lambdatest cli.
 @Library("kautils")
 // Classes we use, under jenkins-jobs/src/.
@@ -49,11 +49,14 @@ Typically not set manually, but rather by other jobs that call this one.""",
 of the GIT_REVISION, especially if it is a commit rather than a branch.
 Defaults to GIT_REVISION.""",
    "master"
-
+// NOTE: These workers are not the same as the workers used by the Jenkins
+// onWorker config. These are the workers we have available in LambdaTest.
+// TODO(FEI-4888): Rename this param to make it clear that these are separate
+// workers.
 ).addStringParam(
    "NUM_WORKER_MACHINES",
-   """How many worker machines to use. Max available is 20.""",
-   "20"
+   """How many worker machines to use in LambdaTest. Max available is 30.""",
+   "30"
 
 )
 .addBooleanParam(
@@ -68,8 +71,8 @@ to spin up.""",
 )
 .addStringParam(
    "TEST_RETRIES",
-   """How many retry attempts to use. By default is 3.""",
-   "3"
+   """How many retry attempts to use. By default is 4.""",
+   "4"
 
 ).addChoiceParam(
    "TEST_TYPE",
@@ -91,7 +94,7 @@ REVISION_DESCRIPTION = params.REVISION_DESCRIPTION ?: params.GIT_REVISION;
 currentBuild.displayName = ("${currentBuild.displayName} " +
                             "(${REVISION_DESCRIPTION})");
 
-// We use the build name as a unique identifier for user notifications. 
+// We use the build name as a unique identifier for user notifications.
 BUILD_NAME = "build e2e-cypress-test #${env.BUILD_NUMBER} (${params.URL}: ${params.REVISION_DESCRIPTION})"
 
 // At this time removing @ before username.
@@ -129,17 +132,17 @@ def runLambdaTest() {
    // Determine which environment we're running against, so we can provide a tag
    // in the LambdaTest build.
    def e2eEnv = params.URL == "https://www.khanacademy.org" ? "prod" : "preprod";
-   
+
    def runLambdaTestArgs = ["yarn",
                             "lambdatest",
                             "--cy='--config baseUrl=\"${params.URL}\",retries=${params.TEST_RETRIES}'",
                             "--bn='${BUILD_NAME}'",
                             "-p=${params.NUM_WORKER_MACHINES}",
-                            "--sync=true", 
+                            "--sync=true",
                             "--bt='jenkins,${e2eEnv}'",
                             "--eof"
    ];
-   
+
    dir('webapp/services/static') {
       withEnv(["LT_USERNAME=${lt_username}",
                "LT_ACCESS_KEY=${lt_access_key}"]) {
@@ -164,7 +167,7 @@ def getUserIds(def deployUsernameBlob) {
 
    for (n in match) {
       // We look for possible duplicates as we don't want to notify the main
-      // user twice.  
+      // user twice.
       // NOTE: This can happen when the deployer includes their name in the
       // deploy command (e.g. `@foo` types: `sun queue foo`).
       if (n[1] != mainUser) {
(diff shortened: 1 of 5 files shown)


jobs/e2e-test-cypress.groovy
```diff
@@ -68,8 +71,8 @@ to spin up.""",
 )
 .addStringParam(
    "TEST_RETRIES",
-   """How many retry attempts to use. By default is 3.""",
-   "3"
+   """How many retry attempts to use. By default is 4.""",
+   "4"
```
jandrade, line 75: **note:** I'm increasing the # of retries to see if we get more consistent results for a couple of flaky tests.

jobs/e2e-test.groovy
```diff
@@ -489,13 +491,13 @@ def runLambda(){
              string(name: 'REVISION_DESCRIPTION', value: REVISION_DESCRIPTION),
              string(name: 'DEPLOYER_USERNAME', value: params.DEPLOYER_USERNAME),
              string(name: 'URL', value: E2E_URL),
-             string(name: 'NUM_WORKER_MACHINES', value: params.NUM_WORKER_MACHINES),
+             string(name: 'NUM_WORKER_MACHINES', value: "30"),
```
jandrade, line 494: **note:** Hardcoding this value here so it doesn't conflict with the Python/selenium workers. This number is used separately so LambdaTest uses it for how many machines we need to use in their platform.

vars/onWorker.groovy
```diff
@@ -3,11 +3,19 @@
...
+// TODO(FEI-4888): Modify this value once we turn off the Python e2e tests. We
+// could probably reduce this to the number of workers we use in the
+// `webapp-test` job.
+
+// How many test-workers we run in parallel, by default. This no longer needs to
+// be consistent across jobs, but most jobs using the ka-test-ec2 workers use it
+// by default.
 def defaultNumTestWorkerMachines() {
-   return 20;
+   return 21;
```
jandrade, line 18: **question:** That makes me think if there's another way to modify the `firstinqueue-priming` job to use a separate primed instance/machine. I'm not fully sure how Jenkins labels and jobs IDs play in this context:

https://github.com/Khan/jenkins-jobs/blob/2a003cd76d1fb59685e0d34adf795fcc8c8cf266/jobs/firstinqueue-priming.groovy#L109-L130

cc @csilvers in case you have any thoughts on this.

jobs/e2e-test-cypress.groovy
```diff
@@ -91,7 +94,7 @@ REVISION_DESCRIPTION = params.REVISION_DESCRIPTION ?: params.GIT_REVISION;
 currentBuild.displayName = ("${currentBuild.displayName} " +
                             "(${REVISION_DESCRIPTION})");
 
-// We use the build name as a unique identifier for user notifications. 
+// We use the build name as a unique identifier for user notifications.
```
jandrade, line 97: **note:** a lot of changes in this file (and in `e2e-test`) are formatting changes (adding newlines, removing whitespaces, etc).

jobs/e2e-test-cypress.groovy
```diff
@@ -49,11 +49,14 @@ Typically not set manually, but rather by other jobs that call this one.""",
 of the GIT_REVISION, especially if it is a commit rather than a branch.
 Defaults to GIT_REVISION.""",
    "master"
-
+// NOTE: These workers are not the same as the workers used by the Jenkins
+// onWorker config. These are the workers we have available in LambdaTest.
+// TODO(FEI-4888): Rename this param to make it clear that these are separate
+// workers.
```
csilvers, line 55: Thanks for adding this comment, that's very helpful!

jobs/e2e-test.groovy
csilvers: Makes sense.  I think this will be clearer once you rename NUM_WORKER_MACHINES in e2e-test-cypress to have a different name.

csilvers: (But maybe just don't pass in this value at all, and let the cypress job use its default?)

vars/onWorker.groovy
csilvers: Yeah, I think the right way to do this is to keep defaultNumTestWorkerMachines and change firstinqueue-priming.groovy to something like:
```
).addStringParam(
   "NUM_WORKER_MACHINES",
   """How many worker machines to use.  Note one worker machine is
   reserved for cypress end-to-end test running.""",
   (onWorker.defaultNumTestWorkerMachines() + 1).toString()
```
You may want a more descriptive doc-change.

csilvers (changes requested):
This makes a lot of sense to me!  "Request changes" to change the default-num-worker var back to normal, and just allocate the extra one in the priming job.

jobs/e2e-test.groovy
jandrade: yeah I like that, I'm going to get rid of this call here.

jobs/firstinqueue-priming.groovy
```diff
@@ -110,7 +110,17 @@ def primeWorkers() {
    def jobs = [
        "sync-webapp": { prime() },
    ];
-   for (def i = 0; i < NUM_WORKER_MACHINES; i++) {
+
```
jandrade, line 113: @csilvers I ended up moving the increment logic inside the `primeWorkers` function so we always make sure to include that extra worker (in case someone runs this job manually).

Let me know if you find it ok, or if I should make any further changes.

jobs/e2e-test.groovy
```diff
@@ -489,13 +489,12 @@ def runLambda(){
              string(name: 'REVISION_DESCRIPTION', value: REVISION_DESCRIPTION),
              string(name: 'DEPLOYER_USERNAME', value: params.DEPLOYER_USERNAME),
              string(name: 'URL', value: E2E_URL),
-             string(name: 'NUM_WORKER_MACHINES', value: params.NUM_WORKER_MACHINES),
              booleanParam(name: 'USE_FIRSTINQUEUE_WORKERS', value: params.USE_FIRSTINQUEUE_WORKERS),
-             string(name: 'TEST_RETRIES', value: "3"),
+             string(name: 'TEST_RETRIES', value: "4"),
```
csilvers, line 493: Again, I'd take this out and just use the default.

jobs/firstinqueue-priming.groovy
csilvers: Yeah, I can see argument either way.  Your call!

csilvers (approved):
shipit

PR #87, opened 2023-02-10
Added work for creating cloud run tasks in jenkins

MiguelCastillo:
## Summary:
In order for Cloud Run services to handle traffic for versions that don't exist in Cloud Run we need to create traffic tags. These traffic tags is how we can send traffic to default revisions.  We need to do this when we build webapp and when we deploy znds.

Issue: https://khanacademy.atlassian.net/browse/INFRA-8963

## Test plan:
Im going to kick off a znd build for service-testbed and check the logs.

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -531,11 +531,8 @@ def uploadGraphqlSafelist() {
 def createCloudRunTags(){
    echo("Creating Cloud Run tags.")
    dir("webapp") {
-      try {
-         exec(["deploy/update_cloud_run_tags.py", NEW_VERSION]);
-      } catch (e) {
-         echo("Failed to wait for new version: ${e}");
-      }
+      exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
+            "--modules_to_ignore", SERVICES.join(', ')]);
    }
 }
 
@@ -654,8 +651,10 @@ def deployAndReport() {
       }
       parallel(jobs);
 
-      createCloudRunTags();
-      uploadGraphqlSafelist();
+      parallel([
+         "create-cloud-run-tags": { createCloudRunTags(); },
+         "update-graphql-safelist": { uploadGraphqlSafelist(); }
+      ])
 
       _alert(alertMsgs.JUST_DEPLOYED,
              [deployUrl: DEPLOY_URL,
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -531,11 +531,11 @@ def uploadGraphqlSafelist() {
 def createCloudRunTags(){
    echo("Creating Cloud Run tags.")
    dir("webapp") {
-            // Note: we do not update tags on the modules that we are deploying,
-            // because they get assigned their tag at deploy-time, so doing it
-            // here would be redundant (and would actually conflict).
-            exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
-            "--modules_to_ignore", SERVICES.join(', ')]);
+      // Note: we do not update tags on the modules that we are deploying,
+      // because they get assigned their tag at deploy-time, so doing it
+      // here would be redundant (and would actually conflict).
+      exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
+            "--modules_to_ignore", SERVICES.join(',')]);
    }
 }
 
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -531,7 +531,10 @@ def uploadGraphqlSafelist() {
 def createCloudRunTags(){
    echo("Creating Cloud Run tags.")
    dir("webapp") {
-      exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
+            // Note: we do not update tags on the modules that we are deploying,
+            // because they get assigned their tag at deploy-time, so doing it
+            // here would be redundant (and would actually conflict).
+            exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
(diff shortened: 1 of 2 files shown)


jobs/build-webapp.groovy
```diff
@@ -528,6 +528,18 @@ def uploadGraphqlSafelist() {
 }
 
 
+def createCloudRunTags(){
+   echo("Creating Cloud Run tags.")
+   dir("webapp") {
+      try {
+         exec(["deploy/update_cloud_run_tags.py", NEW_VERSION]);
```
csilvers, line 535: You need to pass in the list of services being deployed as the 'ignore-services' arg.

jobs/deploy-znd.groovy
```diff
@@ -282,6 +282,17 @@ def deployToGatewayConfig() {
    }
 }
 
+def createCloudRunTags(){
+   echo("Creating Cloud Run tags.")
+   dir("webapp") {
+      try {
+         exec(["deploy/update_cloud_run_tags.py", VERSION]);
```
csilvers, line 289: You need to pass in the list of services being deployed here too.  And also get rid of the try/catch.

jobs/build-webapp.groovy
```diff
@@ -528,6 +528,18 @@ def uploadGraphqlSafelist() {
...
+      } catch (e) {
+         echo("Failed to wait for new version: ${e}");
```
csilvers, line 537: I think this message can be updated. :-)

But I think you should get rid of the try/catch entirely.  This seems like a fatal error to me (since it means we can't talk to our cloud-run services once we set-default.)

jobs/build-webapp.groovy
```diff
@@ -642,6 +654,7 @@ def deployAndReport() {
       }
       parallel(jobs);
 
+      createCloudRunTags();
```
csilvers, line 657: You could put in a TODO that it's possible to do these in parallel.  Or even just do them in parallel right now!

csilvers (changes requested):
Looks good, but you need to add in the ignore-lists.

jobs/build-webapp.groovy
MiguelCastillo: yeah that makes sense! will do.

jobs/build-webapp.groovy
MiguelCastillo: ha! yes :D

jobs/build-webapp.groovy
```diff
@@ -528,6 +528,15 @@ def uploadGraphqlSafelist() {
 }
 
 
+def createCloudRunTags(){
+   echo("Creating Cloud Run tags.")
+   dir("webapp") {
+      exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
+            "--modules_to_ignore", SERVICES.join(', ')]);
```
csilvers, line 535: I don't think that will work.  I think you need `SERVICES.join(',')`, with no space after the comma.

jobs/build-webapp.groovy
```diff
@@ -642,7 +651,10 @@ def deployAndReport() {
       }
       parallel(jobs);
 
-      uploadGraphqlSafelist();
+      parallel([
+         "create-cloud-run-tags": { createCloudRunTags(); },
+         "update-graphql-safelist": { uploadGraphqlSafelist(); }
```
csilvers, line 656: This is fine, but I don't think you need to create a closure just to call it!  You can probably do:
```
"create-cloud-run-tags": createCloudRunTags,
"update-graphql-safelist": uploadGraphqlSafelist,
```
But I don't know groovy very well, so ymmv.

jobs/deploy-znd.groovy
```diff
@@ -282,6 +282,14 @@ def deployToGatewayConfig() {
    }
 }
 
+def createCloudRunTags(){
+   echo("Creating Cloud Run tags.")
+   dir("webapp") {
+      exec(["deploy/update_cloud_run_tags.py", VERSION,
+            "--modules_to_ignore", SERVICES.join(', ')]);
```
csilvers, line 289: (Likewise here.)

jobs/build-webapp.groovy
```diff
@@ -528,6 +528,15 @@ def uploadGraphqlSafelist() {
 }
 
 
+def createCloudRunTags(){
+   echo("Creating Cloud Run tags.")
+   dir("webapp") {
+      exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
```
csilvers, line 534:
```suggestion
            // Note: we do not update tags on the modules that we are deploying,
            // because they get assigned their tag at deploy-time, so doing it
            // here would be redundant (and would actually conflict).
            exec(["deploy/update_cloud_run_tags.py", NEW_VERSION
```

csilvers (approved):
I don't think this is quite right yet, but I also don't think we need another go-around before landing, once you address the below.

jobs/build-webapp.groovy
MiguelCastillo: yeah you are right! did it in a rush.

jobs/build-webapp.groovy
MiguelCastillo: Clearly I dont know groovy very well either... So Im following how everything else does it.  For example https://github.com/Khan/jenkins-jobs/blob/62709da2d847bbe1d3dbeebd2668c6b32a939a4c/jobs/deploy-webapp.groovy#L680

jobs/build-webapp.groovy
MiguelCastillo: thanks for the comment

PR #90, opened 2023-02-27
Change most of our cron-based jenkins jobs to send to slack on success.

csilvers:
## Summary:
We had a problem where the `clean_queues` job deleted a taskqueue it
shouldn't have, and it took us a while to track down because nobody
left at the company even realized this job existed!

This PR tries to help with that a little by having most cron-based
jenkins jobs send to slack on success, rather than just on failure.
That way at least people will have a chance of knowing the job exists.

The only cron-jobs that I did not modify are `make-allcheck` and
`find-failing-taskqueue-tasks`, since they send to slack on success in
any case (as part of their job output); and `update-i18n-lite-videos`,
which pings `@cp-support` when it talks to slack, which we don't
want to do on success (it creates a ticket!)

Issue: https://khanacademy.atlassian.net/browse/INFRA-9008

## Test plan:
Fingers crossed

### DIFF
--- a/jobs/clean-queues.groovy
+++ b/jobs/clean-queues.groovy
@@ -37,7 +37,7 @@ onMaster('1h') {
    notify([slack: [channel: '#infrastructure',
                 sender: 'Mr Monkey',
                 emoji: ':monkey_face:',
-                when: ['FAILURE', 'UNSTABLE', 'ABORTED']]]) {
+                when: ['SUCCESS', 'FAILURE', 'UNSTABLE', 'ABORTED']]]) {
       stage("Deleting queues") {
          deleteQueues();
       }
--- a/jobs/update-ownership-data.groovy
+++ b/jobs/update-ownership-data.groovy
@@ -73,7 +73,7 @@ onMaster('2h') {
    notify([slack: [channel: params.SLACK_CHANNEL,
                    sender: 'Testing Turtle',
                    emoji: ':turtle:',
-                   when: ['FAILURE', 'UNSTABLE', 'ABORTED']]]) {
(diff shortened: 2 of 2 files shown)


MiguelCastillo (approved):
👍

jobs/update-i18n-lite-videos.groovy
```diff
@@ -56,7 +56,7 @@ onMaster('23h') {
                    sender: 'I18N Imp',
                    emoji: ':smiling_imp:', emojiOnFailure: ':imp:',
                    extraText: "@cp-support",
-                   when: ['FAILURE', 'UNSTABLE', 'ABORTED']],
+                   when: ['SUCCESS', 'FAILURE', 'UNSTABLE', 'ABORTED']],
```
mpolyak, line 59: I hope this is going to be one of those cases where Slack messages are not parsed just right, otherwise since it has `extraText: "@cp-support"` it might start generating CP support JIRA tickets :)

If that's the case, we'll need to drop the `@cp-support`.

mpolyak: Found a failing example: https://khanacademy.slack.com/archives/C49296Q7P/p1666098440479329

It didn't generate a JIRA, but it would ping the on rotation `@cp-support` folks. So I think we still want to drop `extraText: "@cp-support",` for this reason if it's going to message on SUCCESS.

csilvers: @mpolyak I can go either way -- is it more importan to ping cp-support on failure, or to see the successes?  I'm guessing the former, so I'll just revert this change.

mpolyak: Yeah, I think lets keep it the way it currently is.

PR #91, opened 2023-02-27
Allow znd's to deploy index.yaml, pubsub.yaml, etc.

csilvers:
## Summary:
I just use the same logic as in build-webapp.groovy.

While in the area I got rid of the dispatch_yaml support in
build-webapp.groovy, now that dispatch.yaml is gone.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1677288416167349?thread_ts=1677287993.252749&cid=C01120CNCS0

## Test plan:
groovy jobs/deploy-znd.groovy.

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -467,21 +467,6 @@ def deployPubsubYaml() {
 }
 
 // This should be called from within a node().
-def deployDispatchYaml() {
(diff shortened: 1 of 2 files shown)


MiguelCastillo (approved):
Thanks!

dbraley (commented):
👍🏻

PR #96, opened 2023-03-28
Clean up potential bad github refs before fetching.

csilvers:
## Summary:
Every so often, deploys will fail with a message like this, at
fetch-time:
> 11:40:01  fatal: bad object refs/heads/ai-guid-acts.prompt-updates

This is because there's a branch, in jenkin's webapp repo, named
ai-guid-acts.prompt-updates, but it points to a bad commit.  I'm not
100% sure how this can happen, but probably it's because the branch is
deleted/never-merged/forced-pushed, and then we do a gc, and now the
commit it was pointed to is gone.

In any case, this "dangling" branch causes `git fetch` to fail.

This PR attempts to deal with this by deleting all branches before a
`git fetch`, when it is safe to do so.  I'm conservative about the
places I do it: they're places where we clone a repo if it doesn't
exist, which is evidence these functions don't have any pre-conceived
notions of what branches might be around.  But they catch the places
that matter for the deploy process: merge-branches and safe-sync-to.

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1680028939583289

## Test plan:
Fingers crossed!  I tried to do this in as safe a way possible, but
it's still a bit scary to automatically delete things.

### DIFF
--- a/safe_git.sh
+++ b/safe_git.sh
@@ -92,6 +92,21 @@ simple_fetch() {
     ( cd "$1" && _fetch )
 }
 
+# Clean out all branches but master/main (and the current branch).
+# This useful because sometimes branches can become dangling (due to
+# force-pushes) and that will break git-fetch.  In situations where
+# you don't care about existing branches, this is a way to fix that.
+# Call from the workspace-root.
+# $1: the repo directory
+clean_branches() {
+    (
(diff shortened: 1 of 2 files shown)


yogieric (approved):
LGTM

PR #98, opened 2023-04-05
FEI-4888.1: Switch to Cypress for e2e tests

jandrade:
## Summary

This is a rollout of the Cypress e2e tests. This is the first
step in the rollout, which is to switch the e2e tests to run on the
Cypress infrastructure instead of the Python-Selenium infra.

The next step will be to remove the `e2e-cypress-*` jobs that were
created while we tested this integration in parallel with the
Python-Selenium infra.

Issue: FEI-4888

## Test Plan

- [x] Run the `e2e-test` job on this branch and verify that it passes in
  Jenkins via the "Replay" button.

NOTE: I created two separate Replay jobs to verify that `first` and
`second` e2e tests pass.

1. first-smoke-test: https://jenkins.khanacademy.org/job/deploy/job/e2e-test/47341/console
2. second-smoke-test: https://jenkins.khanacademy.org/job/deploy/job/e2e-test/47343/console

- [x] Run the `firstinqueue-priming` job on this branch and verify that
  it passes in Jenkins via the "Replay" button.

https://jenkins.khanacademy.org/job/deploy/job/firstinqueue-priming/3019/console

PR #99, opened 2023-04-05
FEI-4888.2: Remove Cypress unused jobs

jandrade:
## Summary:

Follow up to #98 that is responsible for removing the Cypress jobs that
we created for development and testing purposes. Now that we have
switched/moved everything to the `e2e-test` job, we can remove the
files/jobs contained in this PR.

- `e2e-test-cypress` was previously referenced in e2e-test, but that reference was removed in https://github.com/Khan/jenkins-jobs/pull/98.
- `e2e-test-cypress-backup` is in case LT is down, but tbh it's unrealistic to run that job as running cypress tests sequentially would take a LOT of time to be completed, so I'm getting rid of it.

NOTE: Doing this in a separate PR to avoid any issues with the current
Jenkins jobs.

Issue: FEI-4888

## Test plan:

N/A


jeresig (approved):
Great! To clarify: these files aren't referenced anywhere, right?

jandrade:
> Great! To clarify: these files aren't referenced anywhere, right?

@jeresig that's right and good question.

`e2e-test-cypress` was previously referenced in `e2e-test`, but that reference was removed in #98.
`e2e-test-cypress-backup` is in case LT is down, but tbh it's unrealistic to run that job as running cypress tests sequentially would take a LOT of time to be completed, so I'm getting rid of it.

I'm going to add this to the PR description.

somewhatabstract (approved):
LGTM

dbraley (approved):

PR #101, opened 2023-04-13
Fix publish worker and TAP jobs

justinmayhew:
## Summary:
We're currently running into an error when attempting to build a publish worker image:

```
WorkflowScript: 45: illegal string body character after dollar sign;
   solution: either escape a literal dollar sign "\$5" or bracket the value expression "${5}" @ line 45, column 6.
      sh("[ $(df -BM --output=avail . | tr -cd 0-9) -gt 1500 ] || docker image prune -af")
```

🤞 that switching to single quotes solves the problem.

Full error: https://jenkins.khanacademy.org/job/deploy/job/build-publish-image/219/console

Issue: none

## Test plan:
Re-run the publish worker job.

### DIFF
--- a/jobs/build-publish-image.groovy
+++ b/jobs/build-publish-image.groovy
@@ -42,7 +42,7 @@ def runScript() {
                           params.GIT_REVISION);
 
    // Prune docker images before building if under 1.5GB of disk space.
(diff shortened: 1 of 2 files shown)


mpolyak (approved):
🤞 `'`

csilvers (approved):
Good call,! I suspect you're right

dbraley (commented):
👍🏻

PR #102, opened 2023-05-01
Use runpytests.sh in our jenkins jobs.

csilvers:
## Summary:
We now have runpytests.sh as a wrapper around runtests.py to properly
handle both python2 and python3.  Let's use it!

Issue: https://khanacademy.atlassian.net/browse/INFRA-9106

## Test plan:
groovy jobs/update-ownership-data.groovy

### DIFF
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -170,7 +170,7 @@ TESTS = [
     [cmd: "testing/flow_test_client.sh -j1 <server>", done: false],
     [cmd: "testing/typecheck_test_client.sh -j1 <server>", done: false],
     [cmd: "testing/kotlin_test_client.sh -j1 <server>", oneAtATime: true, done: false],
-    [cmd: "testing/runtests.py --quiet --jobs=1 --xml <server>", done: false],
+    [cmd: "testing/runpytests.sh --quiet --jobs=1 --xml <server>", done: false],
     [cmd: "testing/lint_test_client.sh -j1 <server> javascript", done: false],
     [cmd: "testing/lint_test_client.sh -j1 <server> python", done: false],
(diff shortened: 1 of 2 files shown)


jobs/webapp-test.groovy
```diff
@@ -499,7 +499,7 @@ def analyzeResults() {
             if (params.MAX_SIZE != "medium") {
                maxSizeParam = " --max-size=${exec.shellEscape(params.MAX_SIZE)}";
             }
-            pythonRerunCommand = "testing/runtests.py${maxSizeParam}"
+            pythonRerunCommand = "testing/runpytests.sh{maxSizeParam}"
```
latteier, line 502: do we still need the `$`?

MiguelCastillo (approved):
Changes look good!  Do we also want to update https://github.com/Khan/jenkins-jobs/blob/a196a9231f920386c74b7f3009a2e93df6370934/jobs/webapp-test.groovy#L57?

csilvers:
> Changes look good! Do we also want to update
> 
> https://github.com/Khan/jenkins-jobs/blob/a196a9231f920386c74b7f3009a2e93df6370934/jobs/webapp-test.groovy#L57
> ?

No, I chose to leave that as is, because runtests.py is where those flags are defined (and used).  runpytests.sh ignores those flags when running python3 tests, so I thought updating the reference might be actively misleading.

MiguelCastillo:
> > Changes look good! Do we also want to update
> > https://github.com/Khan/jenkins-jobs/blob/a196a9231f920386c74b7f3009a2e93df6370934/jobs/webapp-test.groovy#L57
> > 
> > ?
> 
> No, I chose to leave that as is, because runtests.py is where those flags are defined (and used). runpytests.sh ignores those flags when running python3 tests, so I thought updating the reference might be actively misleading.

ah that makes a ton of sense! cool beans.

PR #105, opened 2023-05-09
[🔥AUDIT🔥] Use the proper test-runner for python tests now.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Now that py_test_client.sh is in master, we can call it directly,
rather than having to go through the shim.

Issue: none

## Test plan:
Fingers crossed


dbraley (approved):

IamMediocracy (commented):
👍

PR #106, opened 2023-05-10
Clean up some jenkins rules and scripts, and make deps more explicit.

csilvers:
## Summary:
We are moving toward a world where we won't build python2 deps by
default as part of `make deps`, so I change the one rule that calls
`make deps` to explicitly build the python2 deps (because it needs
them).

While in the area, I cleaned up some other deps-related code:
graphql-gateway no longer needs js deps; those moved to queryplanner
some time ago.

And I also just deleted the update-current-sqlite.sh script, which
hasn't worked for a while now.  We may resurrect parts of it later,
but we'll have to see what the replacement current.sqlite regime looks
like.

Issue: none

## Test plan:
Fingers crossed


weekly-maintenance.sh
```diff
@@ -396,7 +396,7 @@ set -x
 jenkins-jobs/safe_git.sh sync_to_origin "git@github.com:Khan/webapp" "automated-commits"
 jenkins-jobs/safe_git.sh sync_to_origin "git@github.com:Khan/network-config" "master"
 
-( cd webapp && make deps )
+( cd webapp && make deps python_deps )
```
dbraley, line 399: So I understand, right now this is actually redundant as `make deps` still runs the python_deps, but in the near future that will not be the case, correct?

Either way, LGTM, just want to make sure I understand the current and desired future states.

dbraley (approved):
LGTM

weekly-maintenance.sh
csilvers: Yes, that's correct!

PR #107, opened 2023-05-11
Get rid of unused jenkins jobs and scripts.

csilvers:
## Summary:
I found the unused jobs by running this command on jenkins-server, as
the jenkins user:
```
grep -roh 'jobs/[^ ]*.groovy' ~/jobs/*/jobs/*/*.xml | sort -u
```
This shows all the groovy scripts that are run by currently configured
jobs.

I then compared this against what's in github, and removed the three
jobs that were in github but not in a config file.

Then for every file in the top-level jenkins-jobs repo, I ran
`git grep <filename>`, and deleted all the ones with no hits.
That was most of them!

Issue: none

## Test plan:
Fingers crossed, we can always revert if we need to


MiguelCastillo (commented):
jobs/go-codecoverage.groovy is still a thing https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/ ...?  Perhaps @dmcalpin can speak to that one since you did that work.

jobs/go-postprocessing.groovy... Im not sure.  @tomyedwab do you know?  I think you were the last one to touch that thing.

I can approve and hope for the best.  But Im not entirely sure I know how to help verify these are not used besides asking folks about them.

csilvers:
> jobs/go-codecoverage.groovy is still a thing https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/ ...? Perhaps @dmcalpin can speak to that one since he did that work.

Thanks for the careful look!  I did my diff between the grep output and the ls listing by hand, and must have made a mistake.

MiguelCastillo:
> > jobs/go-codecoverage.groovy is still a thing https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/ ...? Perhaps @dmcalpin can speak to that one since he did that work.
> 
> Thanks for the careful look! I did my diff between the grep output and the ls listing by hand, and must have made a mistake.

Its still possible its running but its no longer in use.  So definitely a still a good find.  I know @dmcalpin  was a super giddy cowboy a while ago about it.  🤣   But haven't heard about it in a while.

tomyedwab:
> jobs/go-postprocessing.groovy... Im not sure. @tomyedwab do you know? I think you were the last one to touch that thing.

I don't honestly remember. I feel like all the translations stuff got moved off Jenkins a long time ago, if that helps.

csilvers:
OK, my grep wasn't wrong!  The go-codecoverage job exists, but it doesn't run go-codecoverage.groovy!  https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/configure shows the groovy contents hard-coded into the xml file, instead of using "script from scm" like all the other jobs.

I'll keep go-codecoverage.groovy, but we'll want to fix that if this is a job we actually want to make use of.

mpolyak (approved):
I looked at the translation-related files, and they're good to delete (as Tom said translations-pipeline is not running anything on Jenkins now days):

```
busy_wait_on_dropbox.sh
download-file.sh
jobs/go-postprocessing.groovy
jobs/run-translation-stats.groovy
run_translation_stats.sh
tap_stats_to_csv_script.py
```

dmcalpin:
@MiguelCastillo @csilvers I'm still a "super giddy cowboy" about the code coverage job, and it's still running. I think it would be good to get a feel from the broader eng org to see if anyone cares about it. If not, I can just run coverage for MPP's services locally, or some other way

dbraley (approved):

csilvers:
I have no desire to delete it.  My main desire is to fix up the config so it runs the groovy file that's checked into source control, rather than having the text part of the config.  Do you mind if I make that change?

dmcalpin:
Yep, that would be fine, but update the repo with what's here first: https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/configure. since over the months I have made some bug fixes to the script.

MiguelCastillo:
> OK, my grep wasn't wrong! The go-codecoverage job exists, but it doesn't run go-codecoverage.groovy! https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/configure shows the groovy contents hard-coded into the xml file, instead of using "script from scm" like all the other jobs.
> 
> I'll keep go-codecoverage.groovy, but we'll want to fix that if this is a job we actually want to make use of.

yeah fixing that is a good idea. thanks for catching!

PR #108, opened 2023-05-11
[🔥AUDIT🔥] Only run `deploy-gateway-config` in the queryplanner directory now.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
We moved this rule fraom graphql-gateway/ to queryplanner/ some time
ago.  We tried to run in both directories for builds of commites
before that migration.  But the migration was like over a year ago
now!  I don't think we need to support the old location anymore.

Issue: https://jenkins.khanacademy.org/job/deploy/job/build-webapp/39021/execution/node/198/log/

## Test plan:


dbraley (commented):
👍🏻

PR #109, opened 2023-05-12
ZNDs:skip deploy-gateway-config on graphql-gateway

yogieric:
We recently changed this for deploying webapp and need to change it for ZNDs too.

Issue: NA

Test Plan:

Fingers crossed, we can always revert if we need to

### DIFF
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -327,11 +327,9 @@ def deployCronYaml() {
 // just do it always.
 def deployToGatewayConfig() {
    dir("webapp") {
-       if (fileExists("services/queryplanner/Makefile")) {
-          exec(["make", "-C", "services/queryplanner",
-                "deploy-gateway-config",
-                "DEPLOY_VERSION=${VERSION}"]);
-       }
+       exec(["make", "-C", "services/queryplanner",
+             "deploy-gateway-config",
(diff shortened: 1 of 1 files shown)


jobs/deploy-znd.groovy
```diff
@@ -327,9 +327,6 @@ def deployCronYaml() {
 // just do it always.
 def deployToGatewayConfig() {
    dir("webapp") {
-       exec(["make", "-C", "services/graphql-gateway",
-             "deploy-gateway-config",
-             "DEPLOY_VERSION=${VERSION}"]);
        if (fileExists("services/queryplanner/Makefile")) {
```
IamMediocracy, line 330: doesn't this if also need to disappear (just the if, not the code in the if) similar to #108 ?

yogieric: I wasn't sure, but Craig pointed it out too. It is removed.

yogieric: Oops... something went wrong with my second commit. Huh.

yogieric: NVM, it is fixed now.

IamMediocracy (approved):
Seems consistent with the other changes that have already been made 👍 
Thanks for taking care of this!

csilvers (commented):
:+1:

PR #110, opened 2023-05-15
[🔥AUDIT🔥] Update the groovy file to match what we're actually running.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
The go-codecoverage jenkins job is currently set up to run groovy in
config.xml, not in go-codecoveage.groovy.  This is the first step to
changing it to use the latter: getting the groovy file to match the
current contents of config.xml, taken from
   https://jenkins.khanacademy.org/job/misc/job/go-codecoverage/configure

Issue: none

## Test plan:
None


dbraley (commented):
👍🏻

PR #113, opened 2023-06-10
INFRA-9290 Log High level jenkins actions to GOOGL

yogieric:
This code logs the start and completion of all onWorker and onMaster steps to the google logging service.  This is useful for debugging and auditing.

All code is meant to be very defensive such that if the logging fails, the build will continue. There is risk that defensive coding was overlooked somewhere and the Khan Academy build system could be impacted. A quick revert is recommended if destabilization is noticed.

It is NOT meant to be efficient. This can likely be fixed with more effort, but the initial goal is to get better telemetry for other efforts. The downside is that this logging should not be used liberally or it could impact performance and cost.

Issue: https://khanacademy.atlassian.net/browse/INFRA-9290

Test Plan:

I've been testing this code using the jenkins replay feature for one webapp-test job. I've also created intentional exceptions to improve my confidence in the defensive coding.

My recommendation is to deploy this code and then watch all active jobs for 60 minutes watching for exceptions (in jenkins logs). If that goes well, keeping an eye on this for the next few days would be good.

### DIFF
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -478,6 +478,7 @@ def analyzeResults() {
          // No need to report the results in the case of abort!  They will
          // likely be more confusing than useful.
          echo('We were aborted; no need to report results.');
+         notify.log("Aborted", ["job":"webapp-test"])
          return;
       }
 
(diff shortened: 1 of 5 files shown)


IamMediocracy (approved):
Agreed with @daelious ! Thank you for making this better!

PR #114, opened 2023-06-12
Update weekly-maintenance.sh to also clean ka-content buckets.

csilvers:
## Summary:
Similar to the ka-translations bucket, we have a lifecycle rule for
these buckets that is too complicated for the built-in google rules.
So we write our own rule in code.

Issue: https://khanacademy.slack.com/archives/C49296Q7P/p1686584563057689

## Test plan:
I temporarily changed the `gsutil rm` line to `echo gsutil rm`, and
ran this function on my local machine, and it emitted lines like:
```
Deleting obsolete snapshot/manifest file gs://ka-content-data/bg/snapshot-1104a1c86e8c5073925bdb8977027830f340bced.json
```

It's hard to tell for certain that it's following the rules exactly,
but I did check that for `am` it did not delete
manifest-c01faf2882146fd73bbee66bc707debbdffc2fa0.json, which is the
10th oldest manifest.  But did delete
manifest-a17244a8df1d97318dc3eaf2e3fb1c7450348d9b.json, which is older
than that.

### DIFF
--- a/weekly-maintenance.sh
+++ b/weekly-maintenance.sh
@@ -174,6 +174,36 @@ clean_ka_translations() {
     done
 }
 
+clean_ka_content_data() {
+    for dir in `gsutil ls gs://ka-content-data gs://ka-revision-data | grep -v :$`; do
+        # This sorts by date.  Each line looks like:
+        #  <size>  YYYY-MM-DDTHH:MM:SSZ  gs://<filename>
+        files=`gsutil ls -l $dir | sort -k2 | grep -v ^TOTAL:`
+
+        # We keep all snapshot/manifest files that fit either of these criteria:
+        # 1) One of the last 10 snapshots/manifests
+        # 2) snapshot/manifest was created within the last 60 days
+        # Based on:
+        # https://khanacademy.slack.com/archives/C49296Q7P/p1686587762692149?thread_ts=1686584563.057689&cid=C49296Q7P
+        # Because we have both manifest and snapshot files in each dir,
+        # we keep the most recent 20 files; that's the 10 most recent uploads.
+        not_last_ten=`echo "$files" | tac | tail -n+20`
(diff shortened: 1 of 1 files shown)


weekly-maintenance.sh
```diff
@@ -174,6 +174,36 @@ clean_ka_translations() {
...
+        # https://khanacademy.slack.com/archives/C49296Q7P/p1686587762692149?thread_ts=1686584563.057689&cid=C49296Q7P
+        # Because we have both manifest and snapshot files in each dir,
+        # we keep the most recent 20 files; that's the 10 most recent uploads.
+        not_last_ten=`echo "$files" | tac | tail -n+20`
+        sixty_days_ago_time_t=`date -d "-60 days" +%s`
+        echo "$not_last_ten" | while read size date fname; do
+            time_t=`date -d "$date" +%s`
+            if [ "$time_t" -lt "$sixty_days_ago_time_t" ]; then
+                # Very basic sanity-check: never delete files from today!
+                if echo "$date" | grep -q `date +%Y-%m-%d`; then
```
mpolyak, line 196: Maybe worth also asserting that it's never `https://www.khanacademy.org/_fastly/published-content-version/<ka locale>` version? That way if the date checks go wonky, we at least prevent the current version for the locales from being deleted.

mpolyak (commented):
Thanks for adding this scary functionality! I wonder if we should first roll this out to only delete `lol` locale files, to ensure it does as intended, while just logging for the other locales?

As well, I think adding a check to ensure we never delete the current published version is also a good safety check to have.

mpolyak (approved):
Thank you for the bash magic!

dbraley (commented):
Ok, yes this LGTM.

IamMediocracy (commented):
Awesome! Thanks for handling yet more cleanup!

daelious:
Thank you! Looks great.

PR #116, opened 2023-06-29
Introducing succeeded, failed, and aborted to monitoring messaging

IamMediocracy:
## Summary:
Previously monitoring was only discerned as "finished" making it difficult
for buildmaster to determine whether there were monitoring issues or not.
This should now provide some better, still not perfect, idea to buildmaster.

Altering a key in notify.groovy to provide consistency between jenkins and buildmaster

Issue: "none"

Test plan:

Run a deploy that succeeds, fails monitoring, and where the monitoring stage is aborted.
This is to some degree :fingers_crossed: as the only place to test this is in prod


jobs/deploy-webapp.groovy
```diff
@@ -615,25 +615,28 @@ def _monitor() {
...
-               // We want to continue with the abort process.
+               // We want to continue with the abort process, while notifying buildmaster.
+               buildmaster.notifyMonitoringStatus(params.GIT_REVISION, "finished:aborted");
                throw e;
-            }
+            } else {
             // Failure to monitor is not a fatal error: we'll tell
             // people on slack so they can monitor manually.  But
             // we don't want to abort the deploy, like a FAILURE would.
+               buildmaster.notifyMonitoringStatus(params.GIT_REVISION, "finished:failed");
```
latteier, line 631: Just so I'm clear - this means that the monitoring process couldn't complete due to an internal failure. not that the monitoring process detected elevated levels of errors, right?

IamMediocracy: I believe you're right about this; I think there's a change to be made to monitor.py that should let it notify buildmaster directly when errors occur... or as Eric previously suggested, that buildmaster consume the monitoring directly. I think it would be easier to have buildmaster and monitor.py talk if the buildmaster token is present than it would be to replace monitor.py with buildmaster handling monitoring. Unless @yogieric can tell me that the monitor.py exits with a status other than `0` when errors are present, a cursory look at the monitor.py does not suggest to me that it does.

MiguelCastillo:
closing this PR. I dont think we need it.

PR #117, opened 2023-07-18
Revert "INFRA-9290 Log High level jenkins actions to GOOGL"

csilvers:
## Summary:

We are putting the parallel-deploys project on hold so let's revert
this change, which adds complexity and latency.

Issue: INFRA-9290

## Test plan:
none


MiguelCastillo (approved):
👍  sounds reasonable to me!

latteier:
I want to be clear that I'd like to revisit the deploy queue experience at some point in the future to see if there are ways to give folks a more productive experience, but I think it makes sense to revert this stuff now rather than leave it in a half-finished state.

PR #121, opened 2023-07-27
[🔥AUDIT🔥] Fix the flags passed to runtests.sh.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Go does not support flags like `-j1`.  It needs to be `-j 1`.

Issue: https://khanacademy.slack.com/archives/C02NMB1R5/p1690449539319869

## Test plan:
Fingers crossed, will try running make-allcheck again


dbraley:
LGTM. I believe you can also do `-j=1` fwiw, but either works for me.

PR #122, opened 2023-07-29
[🔥AUDIT🔥] Switch to .ts for the result reporting script.

jeresig:
🖍 _This is an audit!_ 🖍

## Summary:
Now that we're on TS, we need to use that, instead!

Issue: XXX-XXXX

## Test plan:
We'll be running the Cypress tests and confirming that it works.


kevinbarabash (approved):
LGTM 🚢

jandrade (approved):

PR #124, opened 2023-08-17
Clean up all old ka-content buckets.

csilvers:
## Summary:
To support a gradual rollout, in #114 I made it so we only clean the
lol ka-content bucket.  That seems to have been going well, so now I
extend it to all bucket!

Issue: https://khanacademy.slack.com/archives/C49296Q7P/p1686584563057689

## Test plan:
See above


mpolyak (approved):
🤞

PR #125, opened 2023-08-30
[🔥AUDIT🔥] Log some explanatory text about error messages when deploying.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
prime_query_plan_cache can log some scary-looking errors that are, in
fact, harmless.  Give some explanatory text in the hopes of calming
some of that scariness.

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1693421424702009

## Test plan:

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -505,6 +505,12 @@ def uploadGraphqlSafelist() {
    // the new version.  This depends on the static deploy having
    // finished, since that updates the safelist.
    echo("Pre-computing the query-plans for the latest safelist queries.");
+   echo("NOTE: this command will give a lot of errors like");
+   echo("- Error getting query plan");
+   echo("- query did not validate against schema");
+   echo("This is expected, since some obsolete queries in the safelist do");
+   echo("not validate against the current schema.");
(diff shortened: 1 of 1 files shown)


jobs/build-webapp.groovy
```diff
@@ -505,6 +505,12 @@ def uploadGraphqlSafelist() {
    // the new version.  This depends on the static deploy having
    // finished, since that updates the safelist.
    echo("Pre-computing the query-plans for the latest safelist queries.");
+   echo("NOTE: this command will give a lot of errors like");
+   echo("- Error getting query plan");
+   echo("- query did not validate against schema");
+   echo("This is expected, since some obsolete queries in the safelist do");
+   echo("not validate against the current schema.");
+   echo("These errors do NOT indicate a problem with the deploy.");
```
somewhatabstract, line 513: **question:** Will there ever be errors that look like this that DO mean a problem with the deploy? If so, how will folks differentiate? Can we just not log the ones that aren't a problem if they aren't valuable information?

csilvers: > Will there ever be errors that look like this that DO mean a problem with the deploy?

If there were, it would have shown up in unittests.

PR #126, opened 2023-09-05
Add a job to update the dev-server static images in gcloud.

csilvers:
## Summary:
When called (which we expect to be every day), this job rebuilds
all the images that we use for local dev-server -- one for each
first-party service in webapp -- and uploads them to the google
container registry.  It then updates a file in webapp saying what
git commit corresponds to the latest-pushed images, and checks
that into the `automated-commits` branch.

Issue: https://khanacademy.atlassian.net/browse/INFRA-9151

## Test plan:
`groovy jobs/update-devserver-static-images.groovy` gave only
the expected two errors around `@Library`.

Will try running it and see what happens!

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -55,11 +55,11 @@ def runScript() {
 
 def publishResults() {
    dir("webapp") {
-      sh("git add dev/server/tag-of-latest-uploaded-images.txt");
+      sh("git add dev/server/.env.latest_uploaded_build_tag");
    }
    // Check it in!
    kaGit.safeCommitAndPush(
-      "webapp", ["-m", "Automated update of tag-of-latest-uploaded-images.txt"]);
+      "webapp", ["-m", "Automated update of .env.latest_uploaded_build_tag"]);
 }
 
 
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -1,0 +1,83 @@
+// The pipeline job to update the dev-server static images in gcloud.
+//
+// When called (which we expect to be every day), this job rebuilds
+// all the images that we use for local dev-server -- one for each
+// first-party service in webapp -- and uploads them to the google
+// container registry.  It then updates a file in webapp saying what
+// git commit corresponds to the latest-pushed images, and checks
+// that into the `automated-commits` branch.
+
+@Library("kautils")
+// Classes we use, under jenkins-jobs/src/.
+import org.khanacademy.Setup;
+// Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
+//import vars.exec
+//import vars.kaGit
+//import vars.notify
+//import vars.withTimeout
+
+
+new Setup(steps
+
+).addStringParam(
+   "GIT_BRANCH",
+   """The branch on which to work; we check it out, merge master, and then push
+the updated data-file to it.""",
+   "automated-commits"
+
+).addStringParam(
+   "SLACK_CHANNEL",
+   "The slack channel to send our status info.",
+   "#infrastructure"
+
+).addCronSchedule(
+   '0 2 * * *'
+
+).apply();
+
+
+def runScript() {
+   // We do our work in the automated-commits branch (first pulling in master).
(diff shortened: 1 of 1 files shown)


jobs/update-devserver-static-images.groovy
```diff
@@ -0,0 +1,83 @@
...
+
+
+def publishResults() {
+   dir("webapp") {
+      sh("git add dev/server/tag-of-latest-uploaded-images.txt");
+      // Also publish to GCS, for usage from scripts.
+      // TODO(benkraft): Is this actually the best way for scripts to read this
+      // data?
+      sh("gsutil cp dev/ownership_data.json "
+         + "gs://webapp-artifacts/ownership_data.json");
```
latteier, line 63: I'm confused about why we are updating ownership data here. I wouldn't think that building the images would update ownership data, right?

Is this about combing two nightly jobs into one?

If this is intentional then we should probably mention it in the PR description.

csilvers: Oops, meant to take that line out!

jobs/update-devserver-static-images.groovy
```diff
@@ -0,0 +1,78 @@
...
+}
+
+
+def publishResults() {
+   dir("webapp") {
+      sh("git add dev/server/tag-of-latest-uploaded-images.txt");
+   }
+   // Check it in!
+   kaGit.safeCommitAndPush(
+      "webapp", ["-m", "Automated update of tag-of-latest-uploaded-images.txt"]);
```
dbraley, line 62: Oh! I just realized your other PR is generating this file, not `.env.tags` as I'd previously been using. I kind of prefer the `.env` file since we can just include it and docker knows what to do with it, and it'll still observe what I think most people would expect to be obvious rules around the tags if we want to make it more complicated. That said, this is fine if you prefer this route, it won't be much work for us to just `env BUILD_TAG=$(cat tag-of-latest-uploaded-images.txt)` or whatever on the docker-compose command either.

jobs/update-devserver-static-images.groovy
```diff
@@ -0,0 +1,78 @@
...
+}
+
+
+onMaster('2h') {
+   notify([slack: [channel: params.SLACK_CHANNEL,
+                   sender: 'Devserver Duck',
```
dbraley, line 68: Lol. Love it.

dbraley (approved):
This looks good. Because you're using a different file to update the docker tag than the one I was (sorry, I should have commented that in the ticket) this may require a little additional wiring. If you want to adjust this and the previous PR (https://github.com/Khan/webapp/pull/16458) to use `.env.tags` I feel like that would be better, but if you prefer the txt file we can work with that too. Good either way!

csilvers:
> This looks good. Because you're using a different file to update the docker tag than the one I was (sorry, I should have commented that in the ticket) this may require a little additional wiring. If you want to adjust this and the previous PR ([Khan/webapp#16458](https://github.com/Khan/webapp/pull/16458)) to use `.env.tags` I feel like that would be better, but if you prefer the txt file we can work with that too. Good either way!

Sure, I can change it to use `.env.tags`.  (Can I call it `.env.latest_uploaded_build_tag` instead, or does it have to be `.env.tags`?)

Remind me the format that the file should have.  `BUILD_TAG=<sha>`?  Or just `<sha>`?  Or something else?

dbraley:
> > This looks good. Because you're using a different file to update the docker tag than the one I was (sorry, I should have commented that in the ticket) this may require a little additional wiring. If you want to adjust this and the previous PR ([Khan/webapp#16458](https://github.com/Khan/webapp/pull/16458)) to use `.env.tags` I feel like that would be better, but if you prefer the txt file we can work with that too. Good either way!
> 
> Sure, I can change it to use `.env.tags`. (Can I call it `.env.latest_uploaded_build_tag` instead, or does it have to be `.env.tags`?)
> 
> Remind me the format that the file should have. `BUILD_TAG=<sha>`? Or just `<sha>`? Or something else?

You can use `BUILD_TAG=<sha>` yes. You can also use `BUILD_TAG=BUILD_TAG=dev-static-<yyyy-mm-dd>` if you want. Either will work, so long as the images are tagged that way in the artifact registry, and the `makefile` command will tag it both ways.

PR #127, opened 2023-09-08
[🔥AUDIT🔥] Fix the update-devserver-static-images script after running it in real life.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
One silly mistake, and one not so silly -- I guess the PATH for
jenkins jobs is not what I expected.

Issue: https://jenkins.khanacademy.org/job/misc/job/update-devserver-static-images/4/console

## Test plan:
https://jenkins.khanacademy.org/job/misc/job/update-devserver-static-images/5/console

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -47,8 +47,11 @@ def runScript() {
 
       sh("make fix_deps");  // force a remake of all deps all the time
 
-      // Run the script!
-      sh("Make -C dev/server upload-all-static-images");
+      // Run the script!  We need to add ~/go/bin to the path because
+      // that's where `ko` lives.
+      withEnv(["PATH=${env.PATH}:${env.HOME}/go/bin"]) {
+         sh("make -C dev/server upload-all-static-images");
+      }
    }
 }
 


jobs/update-devserver-static-images.groovy
```diff
@@ -47,8 +47,11 @@ def runScript() {
 
       sh("make fix_deps");  // force a remake of all deps all the time
 
-      // Run the script!
-      sh("Make -C dev/server upload-all-static-images");
+      // Run the script!  We need to add ~/go/bin to the path because
+      // that's where `ko` lives.
```
dbraley, line 51: Weird that `~/go/bin` wasn't on the path already, I thought most go installs did that by default. But yeah, definitely need it!

dbraley (commented):
👍🏻

jobs/update-devserver-static-images.groovy
MiguelCastillo: It is in the `$PATH` as `home/ubuntu/go/bin`...  https://github.com/Khan/aws-config/blob/658cf6bab52476eebf75c1709ab813adb01bf16c/jenkins/.bashrc#L5.  But that path in the build workers at least, is empty.

What's `env.HOME` pointing to?

csilvers: (Not counting all the stuff we add explicitly in `onMaster.groovy`/etc)

MiguelCastillo: We'd have to look closer.  But we have this https://github.com/Khan/aws-config/blob/658cf6bab52476eebf75c1709ab813adb01bf16c/jenkins/setup.sh#L194.  Which should be setting the path...

But perhaps it is what you are saying that groovy is messing with it.  This is what I get in the jenkins server:

```
jenkins@jenkins-server:~$ go env GOPATH
/var/lib/jenkins/go
jenkins@jenkins-server:~$ echo $PATH
/home/ubuntu/webapp-workspace/devtools/khan-linter/bin:/var/lib/jenkins/repositories/khan-linter/bin:/usr/local/google_appengine:/home/ubuntu/google-cloud-sdk/bin:/var/lib/jenkins/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

csilvers: I'm saying I believe that jenkins doesn't get its path from either .profile or .bashrc.  Which makes sense, since it's started by systemd (as root) and not by the jenkins user.

The workers probably do get their path from .profile, since they are started via ssh.  But for master, no.

And indeed:
```
jenkins-master$ sudo ps xfea | grep jenkins
754 ?        Ssl  1469:18 /usr/bin/java -Djava.awt.headless=true -jar /usr/share/java/jenkins.war --webroot=/var/cache/jenkins/war --httpPort=8080 JENKINS_HOME=/var/lib/jenkins USER=jenkins HOME=/var/lib/jenkins NOTIFY_SOCKET=/run/systemd/notify LOGNAME=jenkins JOURNAL_STREAM=8:24836 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin INVOCATION_ID=1367eddd48be42888d7d37aa52669dae LANG=C.UTF-8 SHELL=/bin/bash PWD=/mnt/jenkins
[...]
```
You can see the PATH here is some standard path, not the path set for the jenkins user.

PR #131, opened 2023-09-08
[🔥AUDIT🔥] Add a job to update the dev-server static images in gcloud.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
When called (which we expect to be every day), this job rebuilds
all the images that we use for local dev-server -- one for each
first-party service in webapp -- and uploads them to the google
container registry.  It then updates a file in webapp saying what
git commit corresponds to the latest-pushed images, and checks
that into the `automated-commits` branch.

Issue: https://khanacademy.atlassian.net/browse/INFRA-9151

## Test plan:
`groovy jobs/update-devserver-static-images.groovy` gave only
the expected two errors around `@Library`.

Will try running it and see what happens!

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -55,11 +55,11 @@ def runScript() {
 
 def publishResults() {
    dir("webapp") {
-      sh("git add dev/server/tag-of-latest-uploaded-images.txt");
+      sh("git add dev/server/.env.latest_uploaded_build_tag");
    }
    // Check it in!
    kaGit.safeCommitAndPush(
-      "webapp", ["-m", "Automated update of tag-of-latest-uploaded-images.txt"]);
+      "webapp", ["-m", "Automated update of .env.latest_uploaded_build_tag"]);
 }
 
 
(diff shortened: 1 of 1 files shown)


jobs/update-devserver-static-images.groovy
```diff
@@ -0,0 +1,78 @@
...
+   kaGit.safeMergeFromMaster("webapp", params.GIT_BRANCH);
+
+   dir("webapp") {
+      sh("timeout 2m git clean -ffd"); // in case the merge modified .gitignore
+      sh("timeout 1m git clean -ffd"); // sometimes we need two runs to clean
+
+      sh("make fix_deps");  // force a remake of all deps all the time
+
+      // Run the script!
+      sh("Make -C dev/server upload-all-static-images");
```
dbraley, line 51: Does capital 'M' `Make` work?

dbraley (approved):
I think this is the same as the other job you had, other than the one file name change?

Though I didn't think the `make` command was normally case-insensitive, so maybe a small fix there? But LGTM otherwise (and that's fine if it works too).

jobs/update-devserver-static-images.groovy
csilvers: Nope!  I had to fix that later.

csilvers:
It's unclear to me why this is still open when it's an audit!  And it's definitely been merged.  Weird.

csilvers:
Oh, looks like it was landed in #126.  I have no idea how this dup got created.

PR #132, opened 2023-09-12
[🔥AUDIT🔥] Tell people to run runtests.sh, not runpytests.py.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
The latter is gone now, replaced by the former.  The test-size flag
has also gone away (we don't have any "huge" tests left anymore, so
it's become a non-issue).

Issue: https://khanacademy.slack.com/archives/C02NMB1R5/p1694542433441139

## Test plan:


dbraley (commented):
👍🏻

PR #134, opened 2023-09-13
Make the paths set in onWorker and onMaster match .profile.

csilvers:
## Summary:
Sadly, jenkins doesn't exec `.profile` or `.bashrc` when running
jenkins jobs, it just uses the same path the jenkins binary uses.

On the workers, the jenkins binary (actually the `jar slave.jar`
process) runs via ssh, which *does* exec `.profile`, so all jenkins
jobs on workers run with the path that's in `.profile`.

But on the main jenkins server, the jenkins binary (`jar jenkins.jar`)
runs via systemd, which uses its own default path.  In particular,
systemd runs as root (and later does a proces-chown to the jenkins
user), so it does not use jenkins's .profile.

There might be some way to change this in systemd, but that's pretty
black magic.  Instead, we just set the path we want in our
onWorker.groovy and onMaster.groovy scripts, which run at the
"outside" of all our jenkins jobs.

Unfortunately, the paths set in those scripts had gotten out of date
with the path set in the .profile file.  I get them back in sync, and
add a breadcrumb in .profile so hopefully if the path changes again,
we'll update it everywhere we need to.

This lets me remove the explicit path-adding in
update-devserver-static-images.groovy, which was always an awkward
hack.

Issue: none

## Test plan:
Fingers crossed.  In theory this shouldn't affect any jobs, but we'll
have to land it and see..

### DIFF
--- a/vars/onMaster.groovy
+++ b/vars/onMaster.groovy
@@ -18,9 +18,12 @@ def call(timeoutString, Closure body) {
                 // To export BOTO_CONFIG, for some reason, worker did not
                 // source the .profile or .bashrc anymore.
                 withEnv(["BOTO_CONFIG=${env.HOME}/.boto",
-                         "PATH=/usr/local/google_appengine:" +
+                         "PATH=" +
+                         "/home/ubuntu/webapp-workspace/devtools/khan-linter/bin:" +
+                         "/var/lib/jenkins/repositories/khan-linter/bin" +
+                         "/usr/local/google_appengine:" +
                          "/home/ubuntu/google-cloud-sdk/bin:" +
-                         "${env.HOME}/git-bigfile/bin:" +
+                         "${env.HOME}/go/bin:" +
                          "${env.PATH}"]) {
                      body();
                 }
--- a/vars/onWorker.groovy
+++ b/vars/onWorker.groovy
@@ -62,10 +62,12 @@ def call(def label, def timeoutString, Closure body) {
                // To export BOTO_CONFIG, for some reason, worker did not
                // source the .profile or .bashrc anymore.
                withEnv(["BOTO_CONFIG=/home/ubuntu/.boto",
-                        "PATH=/usr/local/google_appengine:" +
+                        "PATH=" +
+                        "/home/ubuntu/webapp-workspace/devtools/khan-linter/bin:" +
+                        "/var/lib/jenkins/repositories/khan-linter/bin" +
+                        "/usr/local/google_appengine:" +
                         "/home/ubuntu/google-cloud-sdk/bin:" +
-                        "/usr/local/bin/:" +
-                        "${env.HOME}/git-bigfile/bin:" +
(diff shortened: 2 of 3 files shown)


vars/onMaster.groovy
```diff
@@ -18,9 +18,12 @@ def call(timeoutString, Closure body) {
                 // To export BOTO_CONFIG, for some reason, worker did not
                 // source the .profile or .bashrc anymore.
                 withEnv(["BOTO_CONFIG=${env.HOME}/.boto",
-                         "PATH=/usr/local/google_appengine:" +
+                         "PATH=" +
+                         "/home/ubuntu/webapp-workspace/devtools/khan-linter/bin:" +
+                         "/var/lib/jenkins/repositories/khan-linter/bin" +
+                         "/usr/local/google_appengine:" +
                          "/home/ubuntu/google-cloud-sdk/bin:" +
-                         "${env.HOME}/git-bigfile/bin:" +
```
dbraley, line 23: I'm in agreement that git-bigfile _shouldn't_ be in use, but I wonder if it might be safer to leave this for now and remove it in a follow up commit so it can be independently reverted if it's a problem?

vars/onWorker.groovy
```diff
@@ -62,10 +62,12 @@ def call(def label, def timeoutString, Closure body) {
...
                // source the .profile or .bashrc anymore.
                withEnv(["BOTO_CONFIG=/home/ubuntu/.boto",
-                        "PATH=/usr/local/google_appengine:" +
+                        "PATH=" +
+                        "/home/ubuntu/webapp-workspace/devtools/khan-linter/bin:" +
+                        "/var/lib/jenkins/repositories/khan-linter/bin" +
+                        "/usr/local/google_appengine:" +
                         "/home/ubuntu/google-cloud-sdk/bin:" +
-                        "/usr/local/bin/:" +
-                        "${env.HOME}/git-bigfile/bin:" +
```
dbraley, line 68: same comment here about git-bigfile. Also, less sure that we don't use `/usr/local/bin`.

dbraley (approved):
I like this better to be honest, it feels more upfront about the PATH dependencies to declare them in the worker setup, vs just assuming they'll be correct for the user they're run as then munging them when that's wrong.

I do wonder if we should do this in an additive only way for now and remove unused PATH entries in a follow up commit, for the sake of being able to test what we care about in isolation from code cleanup. But if you're confident, I'll trust your judgement.

vars/onMaster.groovy
csilvers: I'm comfortable removing git-bigfile, nothing in this repo mentions the word 'bigfile' anymore after this.

/usr/local/bin is still there -- it's part of the default `$PATH` (which is included here in `env.PATH` on the last line).

MiguelCastillo:
omg - that description illustrated to me why i was having a hard time with what "jenkins" using .bashrc and what not!

PR #136, opened 2023-09-18
FEI-5292: Don't run flow_test_client.sh

kevinbarabash:
## Summary:
We don't use Flow for any of our code in webapp anymore.  This is evidenced by the fact that there are no longer any .flowconfig files in that repo.

NOTE: This should be merged before deploying https://github.com/Khan/webapp/pull/16759.

Issue: FEI-5292

## Test plan:
- n/a


csilvers (approved):
Thank you!

dbraley (commented):
👍🏻

PR #137, opened 2023-09-19
Keep around more of our deploy logs, for longer.

csilvers:
## Summary:
We reap old logs periodically so the jenkins-server disk doesn't fill
up.  But we've been finding sometimes we wish some logs stayed around
a bit longer, for debugging.  This PR makes us save more, to help with
that.

Our jenkins-server has a huge amount of disk space free at the moment
-- something like half a petabyte -- so I'm not worried about logs
filling it up.

Issue: https://khanacademy.slack.com/archives/D9SJ6VASC/p1695073769786779

## Test plan:
groovy jobs/*.groovy

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -46,6 +46,11 @@ new Setup(steps
 // We only need to lock out for promoting; builds are fine to do in parallel.
 ).allowConcurrentBuilds(
 
+// Sometimes we want to debug a deploy a week or two later.  Let's
+// keep a lot of these.
+).resetNumBuildsToKeep(
+   500,
+
 ).addStringParam(
     "GIT_REVISION",
     """<b>REQUIRED</b>. The sha1 to deploy.""",
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -15,7 +15,7 @@ new Setup(steps
 // We run this job once every few minutes; 100 builds covers about
 // 30 minutes.  Let's keep at least a days' around, for debugging.
 ).resetNumBuildsToKeep(
-   5000,
+   9000,
 
 ).addStringParam(
     "VERSION",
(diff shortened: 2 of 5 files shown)


jobs/delete-version.groovy
```diff
@@ -15,7 +15,7 @@ new Setup(steps
 // We run this job once every few minutes; 100 builds covers about
 // 30 minutes.  Let's keep at least a days' around, for debugging.
 ).resetNumBuildsToKeep(
-   5000,
+   9000,
```
MiguelCastillo, line 18: so based on this 100 builds a day, assuming that we have about an 11 hr work day covering east to west coast (not taking into account folks working interesting work hours)...  If we have roughly 200 builds an 1 hour, 9000 builds to keep will roughly translate to 4 work days worth logs?  That seems good to me.

jobs/build-webapp.groovy
```diff
@@ -46,6 +46,11 @@ new Setup(steps
 // We only need to lock out for promoting; builds are fine to do in parallel.
 ).allowConcurrentBuilds(
 
+// Sometimes we want to debug a deploy a week or two later.  Let's
+// keep a lot of these.
+).resetNumBuildsToKeep(
+   500,
```
MiguelCastillo, line 52: This should give us a good window of logs we keep around, which is great.  I wonder if we should copy logs for failed build/tests/deploy somewhere where we dont delete them.  That should hopefully help us keep the likely more interesting logs longer since its not as many of them.

MiguelCastillo (approved):
Awesome, thanks!

jobs/build-webapp.groovy
csilvers: Ooh, interesting idea!  I don't know how to move logs in ways that are useful to view them still.

Another approach would be to mark all failed builds as "keep forever", and then have a cronjob that unkeeps old-enough failed jobs after a while.  But then if folks manually mark a build as "keep forever" then that won't be respected.

MiguelCastillo: oh!  that's actually a really idea too! and probably easier than saving files somewhere...  Cool - I created this ticket https://khanacademy.atlassian.net/browse/INFRA-9507.  I think it's worth while doing.

PR #139, opened 2023-10-17
[🔥AUDIT🔥] Logging for failed http requests to buildmaster

MiguelCastillo:
🖍 _This is an audit!_ 🖍

## Summary:
We are having issues aborting jobs and I have traced it back to a race condition between the http request we send to buildmaster for `prompt-status` for `set-default` and when the message for aborting a job is processed by jenkins.  Its unclear to me what error it is we are getting when the request fails so I am adding logging for it.

## Test plan:
groovy vars/*.groovy


dbraley (commented):
👍🏻

PR #140, opened 2023-10-17
Get rid of the send-to-bugtracker option in jenkins notify().

csilvers:
## Summary:
It was only being used in one place, and as far as I can tell that one
place never worked.  Which is good: the allcheck job hasn't passed
in a year and a half, and we don't need a jira task created for that
fact every night.  The existing slack notifications are more than
sufficient.

So let's get rid of it, not just for cleanliness but
so I can clean up the secrets-requirements of jenkins a little bit.

Issue: none

## Test plan:
Fingers crossed

### DIFF
--- a/jobs/make-allcheck.groovy
+++ b/jobs/make-allcheck.groovy
@@ -63,11 +63,7 @@ onMaster('5h') {
    // individually.
    // TODO(csilvers): remove onMaster(), and just allocate
    // the executor in the notify clean-up steps.
-   notify(
-      [slack: [channel: '#1s-and-0s',
-               when: ['STARTED', 'ABORTED']],
-       bugtracker: [project: 'Infrastructure',
-                    when: ['FAILURE']]]) {
+   notify([slack: [channel: '#1s-and-0s', when: ['STARTED', 'ABORTED']]]) {
       // We need this only to get the secrets to send to slack/asana/etc
       // when there are failures.
       // TODO(csilvers): move those secrets somewhere else instead.
--- a/vars/notify.groovy
+++ b/vars/notify.groovy
@@ -88,8 +88,8 @@ def _statusText(status, includeNumConsecutiveFailures=true) {
 
 
 // Returns shared alertlib requirements, including severity, subject,
-// and body text. Individual sendTo functions (slack, bugtracker, email,
-// and alerta) can build upon these, as needed.
+// and body text. Individual sendTo functions (slack, email, github)
+// can build upon these, as needed.
 def _dataForAlertlib(status, extraText) {
    // Potential additions to the subject may include currentBuild.displayName
    // and env.BUILD_URL. These may be added in the individual service's sendTo.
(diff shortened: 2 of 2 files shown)


jobs/make-allcheck.groovy
```diff
@@ -63,11 +63,7 @@ onMaster('5h') {
    // individually.
    // TODO(csilvers): remove onMaster(), and just allocate
    // the executor in the notify clean-up steps.
-   notify(
-      [slack: [channel: '#1s-and-0s',
-               when: ['STARTED', 'ABORTED']],
-       bugtracker: [project: 'Infrastructure',
-                    when: ['FAILURE']]]) {
```
nathanjd, line 70: It looks like this is failing due to bad Jira authorization? Should we fix it instead of removing it?

```
02:49:14  + echo misc/make-allcheck failed: See https://jenkins.khanacademy.org/job/misc/job/make-allcheck/1302/flowGraphTable/ for what failed (look for the red circle(s)).
02:49:14  
02:49:14  + jenkins-jobs/alertlib/alert.py --severity=error --summary=misc/make-allcheck failed --bugtracker=Infrastructure --cc= --bug-tags=
02:49:14  ERROR:root:Unable to verify Jira project key. This issue may not be created successfully.
02:49:15  ERROR:root:Failed sending None to Jira: HTTP Error 400: Bad Request
02:49:15  ERROR:root:Failed testing current issue for uniqueness. This issue might be created as a duplicate.
02:49:15  ERROR:root:Failed sending {"fields": {"description": "misc/make-allcheck failed: See https://jenkins.khanacademy.org/job/misc/job/make-allcheck/1302/flowGraphTable/ for what failed (look for the red circle(s)).", "issuetype": {"id": "10103"}, "labels": ["auto_generated"], "priority": {"id": "3"}, "project": {"key": "INFRA"}, "reporter": {"name": "jirabot"}, "summary": "misc/make-allcheck failed"}} to Jira: HTTP Error 401: Unauthorized
```

https://jenkins.khanacademy.org/job/misc/job/make-allcheck/1302/console

Though it It looks like the make-allcheck failure was triggered by e2e-test being unable to communicate with buildmaster which @MiguelCastillo seems to be actively looking into with additional logging. So maybe we have enough visibility into these failures already and don't need the Jira INFRA tasks that this is supposed to create? If not, perhaps we change to: `[slack: [channel: '#1s-and-0s', when: ['STARTED', 'ABORTED', 'FAILED']]]`?

```
02:22:42  [ERROR] Buildmaster request failed [resource:commits, params:[git_sha:master, job:second-smoke-test, result:failed, id:53293], SEND_SLACK_COUNT:2, error:hudson.AbortException: Fail: Status code 422 is not in the accepted range: 100:399 while calling https://buildmaster.khanacademy.org/commits, buildmaster_deploy_id:, build_url:https://jenkins.khanacademy.org/job/deploy/job/e2e-test/53293/, build_user:null, build_number:53293, git_revision:master, description:, job_name:deploy/e2e-test, virtual_env:/home/ubuntu/webapp-workspace/env, home:/home/ubuntu, build_status:starting]
```

https://jenkins.khanacademy.org/job/deploy/job/e2e-test/53293/console

csilvers: > It looks like this is failing due to bad Jira authorization? Should we fix it instead of removing it?

A good point.  I've updated the PR message on why it's better just to get rid of it.

> Though it It looks like the make-allcheck failure was triggered by e2e-test being unable to communicate with buildmaster which @MiguelCastillo seems to be actively looking into with additional logging.

https://jenkins.khanacademy.org/job/misc/ shows that `make allcheck` hasn't succeeded in over a year.

> So maybe we have enough visibility into these failures already and don't need the Jira INFRA tasks that this is supposed to create? If not, perhaps we change to: [slack: [channel: '#1s-and-0s', when: ['STARTED', 'ABORTED', 'FAILED']]]?

This job just delegates to two other jobs (one for tests and one for e2e-tess) that already post to 1s-and-0s.  So having this job post there as well is redundant.  We really only need this job to post when it fails to start.  I'll add comments to the code to document this there.

csilvers: > I'll add comments to the code to document this there.

Oh!, it's already there:
>    // We want to notify that make-allcheck started, but don't need to
>   // notify how it did because the sub-jobs will each notify
>   // individually.

nathanjd (approved):
Thanks for the extra info!

dbraley (approved):

PR #141, opened 2023-10-17
Reduce how much we depend on secrets.py in jenkins-jobs.

csilvers:
## Summary:
We're working to get rid of secrets.py (at least once the SAT product
goes away, and we no longer have any python2 in prod).  One large
remaining client is jenkins, which uses `withSecrets()` in a lot of
places.  This PR works to normalize those uses.

It does this in 2 ways:

1. Some uses are unnecessary: they either call python code that
   doesn't require secrets, or they don't even call python code at all
   anymore.  I just removed those withSecerts calls.

2. Some uses are just for slack: they either call a python script that
   only uses secrets to talk to slack, or they call alertlib directly,
   to talk to slack.  In those cases, I use a new
   withSecrets.forSlackAlertlib only.  Right now that is just an alias
   for `withSecrets`, but in a follow-up PR I'll make it fetch just
   the one secret (and directly from GSM, not from secrets.py).

We're left with withSecrets being used in only two places.  The first,
when deploying the SAT product, is unavoidable but will go away once
we no longer deploy python2.  The second is for a few secripts that
are obsolete but still in source control.  We should delete the
scripts, but not in this PR.

Issue: none

## Test plan:
Fingers crossed: I'll try doing a deploy with this on a friday.

### DIFF
--- a/vars/withSecrets.txt
+++ b/vars/withSecrets.txt
@@ -4,3 +4,7 @@ It must be called while the current directory is the workspace-root.
 
 NOTE(benkraft): withSecrets is leaky; other branches of the same job will see
 the secrets file too, even if they are not in a withSecrets block.
+
+We also provide the ability to just decrypt one secret: the slack
+secret that alertlib needs to talk to slack.  It is a surprisingly
+common case that that is the only secret we need for a given use.
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -262,7 +262,7 @@ def _alert(def slackArgs, def interpolationArgs) {
                                             interpolationArgs);
       args += ["--slack-attachments=${attachmentString}"];
    }
-   withSecrets() {     // to talk to slack
+   withSecrets.slackAlertlibOnly() {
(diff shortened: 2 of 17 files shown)


vars/withSecrets.txt
```diff
@@ -4,3 +4,7 @@ It must be called while the current directory is the workspace-root.
 
 NOTE(benkraft): withSecrets is leaky; other branches of the same job will see
 the secrets file too, even if they are not in a withSecrets block.
+
+We also provide the ability to just decrypt one secret: the slack
```
nathanjd, line 8: Should this also be a TODO?

nathanjd (approved):
I'm always a fan of these types of simplifications.

vars/withSecrets.txt
csilvers: I see it as just describing the function that only decrypts one secret, which actually is part of this PR.  The fact the function does extra work isn't really relevant to this help text...

dbraley (approved):

PR #142, opened 2023-10-17
Add a `when` to the github notification type.

csilvers:
## Summary:
This makes it consistent with all the other types.  Since we want to
always send to github, the `when` is all possible states.  (Though
maybe we don't want it for aborted?)

Issue: https://khanacademy.slack.com/archives/C013ANU53LK/p1697576502336959

## Test plan:
Will test on jenkins


jeresig (approved):

dbraley (commented):
👍🏻 Makes sense to me.

PR #145, opened 2023-10-20
Provide the ability to only load only a few secrets for calling alertlib.

csilvers:
## Summary:
Now, when you send to slack or github using alertlib, we only load the
one slack secret (or github secret) needed for this.  And we load it
directly from GSM, rather than depending on secrets.py.

As a happy side-effect, this means we can now use use alertlib without
needing webapp cloned.  I take advantage of that by no longer cloning
webapp in the make-allcheck job, which we used to do just to get
secrets for sending alerts to slack.

Issue: none

## Test plan:
Will try running on jenkins!


dbraley (approved):
Cool, works for me!!

PR #146, opened 2023-10-20
Get rid of send-to-email support, which hasn't worked in a long time.

csilvers:
## Summary:
Looking over the alertlib code, I noticed that the secrets that
alertlib needs to send email -- sendgrid_username and
sendgrid_password -- do not exist in secrets.py, and possibly haven't
for a long time.  This means that sending to email would never work.
Given that, in jenkins, we only use send-to-email mode for two jobs
and then only on failure, and given that those jobs haven't failed in
months and months, it's possible that this feature hasn't worked for a
long time and we never noticed.  (Especially since the email goes to
jenkins-admin, which as far as I know I'm the only one who ever looks
at it.)

So let's just get rid of this feature, and depend to sending to slack
on failure.  That seems to have been working well so far.

I considered trying to fix send-to-email instead, by restoring these
sendgrid secrets to secrets.py.  But I couldn't find the secrets in
either gsm or keeper.  Rather than try to track them down or do other
heroic efforts to revive this feature, I think it's better just to let
it Rest In Peace.

Issue: none

## Test plan:
Will run on jenkins.

### DIFF
--- a/jobs/update-i18n-lite-videos.groovy
+++ b/jobs/update-i18n-lite-videos.groovy
@@ -56,9 +56,7 @@ onMaster('23h') {
                    sender: 'I18N Imp',
                    emoji: ':smiling_imp:', emojiOnFailure: ':imp:',
                    extraText: "@cp-support",
-                   when: ['FAILURE', 'UNSTABLE', 'ABORTED']],
-           email: [to: 'jenkins-admin+builds',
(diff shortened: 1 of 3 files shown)


dbraley (approved):
👍🏻

csilvers:
FWIW, it turns out that we don't actually need sendgrid secrets when running on jenkins; alertlib just uses sendmail in that case.  So I reverted this PR (in 1f5c2e6f288bf4055a4278e1e0c2f01924fd4c49).

PR #147, opened 2023-10-20
[🔥AUDIT🔥] Add back in secrets in a place where they're needed.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
We need the graphql-safelist-secret to upload to graphql.

I also rewrote withSecrets in a way that should not leak the secrets
in the jenkins logs anymore.

Issue: INFRA-XXXX

## Test plan:


dbraley (commented):
👍🏻

PR #148, opened 2023-10-20
[🔥AUDIT🔥] Don't decrypt secrets for uploading the graphql-safelist.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
I rewrote the deploy script to get the secret directly from gsm, so it
doesn't need secrets.py.

Issue: None

## Test plan:
Will do a test deploy on jenkins


dbraley (commented):
👍🏻

PR #152, opened 2023-10-20
[🔥AUDIT🔥] Decrypt both slack and stackdriver secrets, in two places.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
There are a couple of scripts we run that send data to stackdriver:
monitoring and set-default.  Let's make sure they have the secrets
they need.

While writing this I discovered a bug in how we were accounting for
how the secrests blocks were being used, meaning that we weren't
deleting secrets-files when we were done with them, like we were
supposed to.  I fixed that up too.

Issue: None

## Test plan:
Will try a deploy

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -572,7 +572,10 @@ def _promoteServices() {  // call from webapp-root
     }
 
     cmd += [services.join(",")];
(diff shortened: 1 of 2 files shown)


dbraley (commented):
👍🏻

PR #154, opened 2023-10-20
[🔥AUDIT🔥] Create our secrets.py files atomically.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
We were having an issue where multiple jobs running in parallel all
tried to create a secrets.py file at the same time.  Since we create
them incrementally, they were stepping on each other, causing
problems.

I solve that not by locking, but by letting each job create its own
secrets.py file, and then using an atomic `mv` file to create the
final secrets.py.  So it should always have a valid value.

Issue: none

## Test plan:
Will try a deploy

### DIFF
--- a/vars/withSecrets.groovy
+++ b/vars/withSecrets.groovy
@@ -51,8 +51,11 @@ def call(Closure body) {
 def slackAlertlibOnly(Closure body) {
    try {
       sh("mkdir -p decrypted_secrets/slack/");
-      sh("gcloud --project khan-academy secrets versions access latest --secret Slack__API_token_for_alertlib >decrypted_secrets/slack/secrets.py");
-      sh("perl -pli -e 's/^/slack_alertlib_api_token = \"/; s/\$/\"/' decrypted_secrets/slack/secrets.py");
+      sh("gcloud --project khan-academy secrets versions access latest --secret Slack__API_token_for_alertlib >decrypted_secrets/slack/secrets.py.tmp");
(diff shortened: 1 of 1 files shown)


dbraley (commented):
Ok, that took a minute to grok, because there are are 3 `secrets.py` files in this commit, but they're different files entirely. But it's the individual methods that are being parallelized and and stomping on each other I think? IE _n_ instances of `slackAlertlibOnly` happening in parallel all writing to `decrypted_secrets/slack/secrets.py`?

I don't see you creating a per-run unique temp file though? A `*.tmp` file yes, but they'll all use that exact same `*.tmp` file won't they? So either I'm missing something, or I don't think this is going to work.

I probably need more coffee though, sorry.

csilvers:
> Ok, that took a minute to grok, because there are are 3 `secrets.py` files in this commit, but they're different files entirely. But it's the individual methods that are being parallelized and and stomping on each other I think? IE _n_ instances of `slackAlertlibOnly` happening in parallel all writing to `decrypted_secrets/slack/secrets.py`?

That's exactly right.

> I don't see you creating a per-run unique temp file though? A `*.tmp` file yes, but they'll all use that exact same `*.tmp` file won't they? So either I'm missing something, or I don't think this is going to work.

They don't all use the exact same tmp file.  The tmp filename includes the pid, and each call to slackAlertlibOnly will have its own pid.

PR #157, opened 2023-10-20
Access the buildmaster token from GSM, not from a file on disk.

csilvers:
## Summary:
This is part of our continuing effort to get rid of -- or at least
minimize -- secrets at rest on the Jenkins machine.  Once this is
landed, we can get rid of all the foo.secrets files in `~jenkins/` on
jenkins-master!

Issue: none

## Test plan:
Will try to run a deploy


nathanjd (approved):
Good thing GSM isn't rate limiting us like Keeper.

dbraley (approved):
LGTM

PR #159, opened 2023-10-20
[🔥AUDIT🔥] Make our rename of secrets.py more atomic.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
We were doing:
```
rm foo
mv foo.tmp foo
chmod 600 foo
```
which is a race condition because if two jobs are doing that
simultaneously, one can do the `chmod` while the other is between the
`rm` and the `mv`, and then the chmod fails.

I get rid of the `rm`, which wasn't necessary, to avoid this.

Issue: https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/16349/flowGraphTable/

## Test plan:
Will try deploying


dbraley (commented):
👍🏻

PR #161, opened 2023-10-23
Remove svgcrush and pngcrush from weekly-maintenance.

csilvers:
## Summary:
It's not safe to run these anymore, because we added a new lint check
that static files like images do not change their contents after being
deployed.  But compressing them would run the risk of doing exactly
that.  We'll have to re-implement this functionality in a different
way, perhaps as a commit hook (e.g. a github action).

Issue: https://github.com/Khan/webapp/pull/17503#pullrequestreview-1692766676

## Test plan:
None


jeresig (approved):
Great, thanks Craig!

PR #162, opened 2023-10-24
Added logic to detect and alert of potentially stuck deployment jobs

MiguelCastillo:
## Summary:
We are having random deploy jobs getting stuck causing all sorts of nasty issues in the deploy queue with seeming stuck jobs after a user has abort their deployment. The theory is that abort signals from jenkins to threads are not working because the threads are in uninterruptible cycles.  To help us determine this I have added slack alerts to the `#infrastructure-deploys` channel to let us know that we pontentially have deploy-webapp jobs that are stuck.

I have added logic to trigger an abort but I have it disabled to start with to ensure that we have the correct logic in place before potentially start to abort jobs that shouldn't be.  Once this is out and we can test it, we can enable the logic to abort build that we deem "stuck".

## Test plan:
Monitor the deploy and verify that we don't have deploy jobs suddenly aborting.  Icing on the cake, try to reproduce the stuck aborted jobs to test the slack logging logic.

1. queue a deploy.
2. when asked to `set default` by sun, monitor your deploy-webapp job and try to abort when we are making the request to buildmaster to get the `prompt-status`.
3. If the job is stuck, then we will keep looping asking for the prompt-status.
4. Check #infrastucture-deploys channel for informational message about the deploy.

This is pretty tricky to reproduce... And best done when nobody is in the queue.

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -208,6 +208,14 @@ NEW_VERSION = null;
 alertMsgs = null;
 
 
+// Sometimes jobs that are aborted by a user don't really abort because
+// Jenkins job is in an uninterruptible status. Most likely due the abort
+// happened while the job is making an HTTP request. In those case, we will
+// throw this exception to really really abort the jobs to prevent getting
+// stuck with jobs not aborting.
+public class AbortDeployJob extends Exception {}
+
+
 @NonCPS     // for replaceAll()
 def _interpolateString(def s, def interpolationArgs) {
    // Arguments to replaceAll().  `all` is the entire regexp match,
@@ -260,6 +268,21 @@ def _alert(def slackArgs, def interpolationArgs) {
       args += ["--slack-attachments=${attachmentString}"];
    }
    withSecrets.slackAlertlibOnly() {
+      sh("echo ${exec.shellEscape(msg)} | ${exec.shellEscapeList(args)}");
+   }
+}
+
+// TODO(miguel): remove once we are done monitoring hanging deploys.
+def _simpleSlackAlert(def channel, def msg, def severity) {
+   args = ["jenkins-jobs/alertlib/alert.py",
+           "--slack=${channel}",
+           "--chat-sender=Mr Anderson",
+           "--icon-emoji=:crying-nickcage:",
+           "--slack-simple-message",
+           "--severity=${severity}",
+          ];
+   withSecrets.slackAlertlibOnly() {     // to talk to slack
+      echo("sending slack message")
       sh("echo ${exec.shellEscape(msg)} | ${exec.shellEscapeList(args)}");
    }
 }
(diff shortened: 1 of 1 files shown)


jobs/deploy-webapp.groovy
```diff
@@ -264,6 +272,21 @@ def _alert(def slackArgs, def interpolationArgs) {
    }
 }
 
+// TODO(miguel): remove once we are done monitoring hanging deploys.
+def _simpleSlackAlert(def channel, def msg, def severity) {
+   args = ["jenkins-jobs/alertlib/alert.py",
+           "--slack=${channel}",
+           "--chat-sender=Mr Anderson",
+           "--icon-emoji=:crying-nickcage:",
```
csilvers, line 280: :crying_cat_face:

jobs/deploy-webapp.groovy
```diff
@@ -519,13 +542,75 @@ def _manualPromptCheck(prompt){
...
+
    def status;
    while (status != "confirmed") {
+      def buildResult = currentBuild.result
       status = buildmaster.pingForPromptStatus(prompt,
                                                params.GIT_REVISION)
 
+      _maybeAbortJob(
+         buildResult,
+         currentBuild.result,
```
dnerdy, line 604: I think it's pretty unlikely the job state will change will prompting for status. The change is most likely to occur because of a buildmaster or manual abort, which can happen at any time. If the current build is aborted, is the result really "ABORTED"? If so, that's great and it seems like we could just check if the current build has that (like is done in the `if` at the top of the function).

MiguelCastillo, line 604: I believe that the `result` is changed to `ABORTED` but the thread never dies.  That's the assumption I am making.  But Im going to setup a small test to try to reproduce this.

1. setup a rest endpoint somewhere that will respond extremely slowly by adding a sleep once it gets the request.
2. have a test jenkins job call that endpoint.
3. abort the job by calling the `stop` endpoint.
4. Hope to reproduce the issue of the build thread hanging.  Check if `result` is set to `ABORTED` even tho the thread is still running

dnerdy (commented):
I like the idea of handling this in Jenkins. I kind also wonder if this could be handled via some "really abort the job" plugin, though I have no idea how such a plugin would work (or if it's even possible).

jobs/deploy-webapp.groovy
MiguelCastillo: Yes i just left a comment around that in this bit of code.  Jenkins logs do show that even tho the job is stuck, jenkins did try to abort the job.  At some point i was able to reproduce the issue and the call to `api/json` reported an aborted status even tho the build was still looping.  So unless the API is returning something that isn't matching the actual state of jenkins, this should work.   But im going to setup a way to test this more directly.

csilvers (approved):
I like the prudent, slow-and-steady approach here.  I'm interested to see what data comes in.

But is this the thread that normally keeps running even after an abort?  I guess maybe it is in practice, because this is a common time to abort.  But in theory we could see this problem whenever we abort -- it could be in setDefaultAndMonitor, for instance, or I guess in verifyPromptConfirmed, so this fix might not be enough.

I wonder if we need something else like `_watchdog` (from notify.groovy).  It would be helpful to look at a job where the user aborted but the job didn't terminate properly

MiguelCastillo:
> I like the prudent, slow-and-steady approach here. I'm interested to see what data comes in.
> 
> But is this the thread that normally keeps running even after an abort? I guess maybe it is in practice, because this is a common time to abort. But in theory we could see this problem whenever we abort -- it could be in setDefaultAndMonitor, for instance, or I guess in verifyPromptConfirmed, so this fix might not be enough.
> 
> I wonder if we need something else like `_watchdog` (from notify.groovy). It would be helpful to look at a job where the user aborted but the job didn't terminate properly

Yeah definitely - this a discovery change.  Once we can confirm that `result` has the correct state, then we will need a more robust way to handle this.  I didnt want to change _watchdog quite yet tho.  But that's probably the best place to systematically handle this.

csilvers:
Based on https://khanacademy.slack.com/archives/C01120CNCS0/p1698170033750009 I'm thinking we should revert this.  I think the fix will be more reliable if we fix all our `catch`s rather than trying to patch up a bad `catch` after the fact (and in only one place!)

csilvers:
Here's a more concrete suggestion: let's collect data from this change, specifically around `currentBuild.result`.  If we find that currentBuild.result is being set properly -- I think it should be -- then we should go find all the places that do `catch (e)` and have them look at currentBuild.result.  Right now only some of them do (e.g. the one in deploy-webapp.groovy:_promote).

PR #164, opened 2023-10-24
Make sure we check for abort every time we catch an exception.

csilvers:
## Summary:
If a jenkins job is aborted, every process gets passed an
AbortException.  But it's possible to swallow that exception if your
code is in a try/catch.

This PR looks at every `catch` in our codebase, and if the `catch`
doesn't always rethrow anyway, I add a call to a function that checks
if the job has been aborted, and if so rethrows the exception (which
is presumably AbortException).

That should resolve issues like
https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/16256
where a job got aborted but stuff kept on running anyway, because the
`notify.log()` function swallowed the abort exception.

## Test plan:
Am mostly just concerned this doesn't break anything; we'll have to
land it and see.

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -683,13 +683,7 @@ def _promote() {
          // messages as well.)
          buildmaster.notifyDefaultSet(params.GIT_REVISION, "finished");
       } catch (e) {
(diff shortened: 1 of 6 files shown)


MiguelCastillo (approved):
Looks great!  Let's give this a try.  Let's do it!

PR #167, opened 2023-10-26
Hnadle the case where the deploy-username isn't set.

csilvers:
## Summary:
The deploy-username is set when the e2e tests are run by jenkins, but
not when run by hand.  Let's make sure the script gracefully handles
that case.

Issue: https://jenkins.khanacademy.org/job/deploy/job/e2e-test/53606/console

## Test plan:
I replayed job 53606 (as 53607) with this new code, and it passed.


MiguelCastillo (approved):
Makes sense!

PR #168, opened 2023-10-26
[🔥AUDIT🔥] Fix the gcloud account we try to run our network-config backups as.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
We used to use prod-read, but I think that account went away.  Now we
use the service account associated with the existing
gcloud-service-account.json file.

I hope (expect) this service account has all the permissions we need.
In general, we're not very good about limiting permissions for service
accounts.

Issue: none

## Test plan:
Ran
```
make -C stackdriver ACCOUNT=stackdriver-service-account@khan-academy.iam.gserviceaccount.com
```
on jenkins-server in the webapp-maintenance job directory, and got no
access errors.

### DIFF
--- a/weekly-maintenance.sh
+++ b/weekly-maintenance.sh
@@ -297,7 +297,7 @@ backup_network_config() {
     (
         cd network-config
         make deps
-        make ACCOUNT=storage-read@khanacademy.org CONFIG=$HOME/s3-reader.cfg PROFILE=default GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcloud-service-account.json
+        make ACCOUNT=stackdriver-service-account@khan-academy.iam.gserviceaccount.com CONFIG=$HOME/s3-reader.cfg PROFILE=default GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcloud-service-account.json
         git add .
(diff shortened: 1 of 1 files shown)


dbraley (commented):
I'm curious if the `stackdriver-service-account` is the right one to be making this reqeust with? At the very least, it's not clear to me why that is the account we should be using to back up network configuration?

But 👍🏻 anyway, because this is an audit, and more importantly if it works, it's better than it being broken

csilvers:
> I'm curious if the `stackdriver-service-account` is the right one to be making this reqeust with? At the very least, it's not clear to me why that is the account we should be using to back up network configuration?
> 
> But 👍🏻 anyway, because this is an audit, and more importantly if it works, it's better than it being broken

That is the email that is associated with the gcloud-service-account.json that the stackdriver script uses.  I assume they have to match?  It's possible, actually, that once we say to use that json file, it uses the stackdriver-service-account by default and we don't need to specify it explicitly.  This is consistent with what I've seen that historically the stackdriver script only produced output for a few gcloud projects, the ones that stackdriver-service-account has access to, even though prod-read@ had access to many more projects.

PR #169, opened 2023-10-26
[🔥AUDIT🔥] Do not pass the `--deployer` flag when no deployer is known.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
notify-e2e-results.ts is ok with `--deployer` being missing, but
not with it being the empty string.

Issue: https://jenkins.khanacademy.org/job/deploy/job/e2e-test/53611/console

## Test plan:
groovy jobs/e2e-test.groovy

PR #170, opened 2023-10-26
[🔥AUDIT🔥] Make sure jenkins-server runs as prod-deploy@.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
I had activated a service-account in #169, but I didn't realize that
this command makes that service-account the default for all gcloud
operations.  We still want the default to be done as
prod-deploy@khanacademy.org!  And only use the service-account with
the `--account` flag.  This new command does that.

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1698352831330619

## Test plan:
Ran the command manually on jenkins-server with success.

### DIFF
--- a/weekly-maintenance.sh
+++ b/weekly-maintenance.sh
@@ -297,7 +297,7 @@ backup_network_config() {
     (
         cd network-config
         make deps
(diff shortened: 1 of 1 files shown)


dbraley (commented):
👍🏻

PR #171, opened 2023-10-26
Use runlint.sh instead of the lint-test-client for jenkins tests.

csilvers:
## Summary:
This is the version of the test-runner that has been ported to Go!

Issue: none

## Test plan:
Ran this on jenkins with success.


dbraley (approved):
cool!

MiguelCastillo (approved):
nice!

PR #174, opened 2023-10-28
[🔥AUDIT🔥] Finally get the script to actually report failed jobs correctly.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
`set -e` does not behave in a useful way, sadly.  There is no way
(that I found) to have `set -e` fail a function if any command in that
function fails, but then have the caller of the function continue to
run even if the function call fails.  (This is because a `||` or a
`&&` disables the `-e` behavior in all code on the left side of the
operator.)

The only way to get the behavior I want -- that each individual job
runs with `set -e` semantics, but if a job fails we just mark it and
continue to the next job -- is to run the jobs in a separate process.
I can set `-e` in the separate process, while the driver runs without
`-e` and just keeps track of failures itself.

It's arguably easier to follow along what's going on now, though I'm
sad to have to split the code into two files.

Issue: none

## Test plan:
From the parent-dir of this repo, I ran:
```
jenkins-jobs/weekly-maintenance.sh clean_docker
```
with success.

I then changed `clean_publish_worker` to have `false` as its first
command, and ran:
```
jenkins-jobs/weekly-maintenance.sh clean_publish_worker clean_docker
```
and it exited with rc 1, saying:
```
THE FOLLOWING JOBS FAILED:  clean_publish_worker
```
but the clean-docker command still ran (as I could see from the output
of running the script).


weekly-maintenance-jobs.sh
```diff
@@ -0,0 +1,345 @@
...
+# NOTE: weekly-maintenance.sh runs the functions in the order listed
+# in this file, so if the job times out, stuff at the end is least
+# likely to have gotten run.  So think carefully about where you place
+# new functions!  Probably you want to put your new function before
+# `clean_package_files`.
+#
+# Here are some cleanups we'd like to add:
+#   vacuum unused indexes
+
+set -ex
```
dbraley, line 14: Did you mean to leave the 'x' here?

dbraley (commented):
Ok, this makes way more sense to me than the previous Pr that dealt with this same logic. I am kind of assuming that everything that moved was more or less moved without change. Spot check indicated that was correct, other than the addition of the `x` flag, which I suspect got left behind on accident - and i'd wager is probably removed in a follow up PR I haven't looked at yet.

MiguelCastillo:
FWIW - I like this better

PR #175, opened 2023-10-30
[🔥AUDIT🔥] Work around a gsutil bug when cleaning ka-translations.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Sometimes the `gsutil ls <dir>` command lists the directory itself.  I
think that's a bug in gsutil -- it occurs in the gsutil installed with
gcloud 419, but not with gcloud 452 -- but it's easy enough to work
around it, so let's just do that.

Issue: https://jenkins.khanacademy.org/job/misc/job/webapp-maintenance/272/execution/node/264/log/

## Test plan:
Ran the command manually with success.


MiguelCastillo (commented):
Great!

dbraley (commented):
👍🏻

PR #176, opened 2023-10-30
[🔥AUDIT🔥] There's another `raw` directory in ka-translations now.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Issue: https://jenkins.khanacademy.org/job/misc/job/webapp-maintenance/273/consoleFull

## Test plan:
Fingers crossed, will try running the job again.


dbraley (commented):
Is there a reason we wouldn't want to skip all `raw*` directories? Probably early optimization though, I don't know if we'd ever have more we need to consider.

PR #177, opened 2023-10-30
Get rid of the clean_ka_translations job.

csilvers:
## Summary:
The ka_translations bucket has gone away!  We don't use it anymore:
   https://khanacademy.slack.com/archives/C49296Q7P/p1698680412983269

Instead, we use ka_translations_archive, but that has a normal
lifecycle rule, that we can use the normal gcs API for, and we don't
need a special thing in the weekly-maintenance script.

Issue: https://khanacademy.slack.com/archives/C49296Q7P/p1698680412983269

## Test plan:
None


mpolyak (approved):
Thanks!

PR #179, opened 2023-11-28
[🔥AUDIT🔥] Let the update-devserver-images job take longer.  A lot longer.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Now that we're cross-compiling all our Go binaries on arm, using
emulation, it can take like 20 minutes to build the containers for a
single service.  So let's give it 10 hours to finish totally.

(A better solution is to use cross-compilation instead of emulation,
but that is a task for another day.)

Issue: none

## Test plan:
Fingers crossed

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -1,4 +1,4 @@
-// The pipeline job to update the dev-server static images in gcloud.
+// The pipeline job to update the dev-server images in gcloud.
 //
 // When called (which we expect to be every day), this job rebuilds
 // all the images that we use for local dev-server -- one for each
@@ -31,7 +31,7 @@ the updated data-file to it.""",
    "#infrastructure"
 
 ).addCronSchedule(
-   '0 2 * * *'
+   '0 22 * * *'
 
 ).apply();
 
@@ -67,7 +67,7 @@ def publishResults() {
 }
 
 
-onMaster('2h') {
+onMaster('10h') {
    notify([slack: [channel: params.SLACK_CHANNEL,
                    sender: 'Devserver Duck',
                    emoji: ':duck:',


jobs/update-devserver-static-images.groovy
```diff
@@ -31,7 +31,7 @@ the updated data-file to it.""",
    "#infrastructure"
 
 ).addCronSchedule(
-   '0 2 * * *'
+   '0 22 * * *'
```
dbraley, line 34: Good idea to move this up

dbraley (commented):
Yeah, I can't say I'm surprised by that. It's worth noting that if this thing actually takes 10 hours, even with the start pushed back to 10pm PDT, that finishes up at 6am PDT, which is 9am EDT. I know there's a lot of us out here on the east side that are up and working earlier than that.

It's still probably not a big deal, as there's not any real harm from that happening after the east coast starts work, it could be somewhat confusing if someone lands a service change BEFORE the images are minted, and those minted images don't include that service change. Though still that seems pretty unlikely, and would still not be all that terrible, as we tag the images based on the branch they where built from, NOT the commit they land on - so while chronologically it might be confusing, the git history would be quite clear about what was included and what wasn't.

Further, just because we give it 10 hours to succeed doesn't mean it will take that long. Right now, it should just be the two images that are building multi-arch I think, though adding go to that may slow things down further.

csilvers:
Yeah, I'm not sure we will need the full 10 hours; I am doing cross-compilation again in #18312 so we're back to the whole thing taking an hour.  But as we add more images and stuff it will likely take longer, so I feel having a way-too-long timeout is fine.  We just need the timeout to be less than 24 hours, so we don't have two of these running at once!

PR #180, opened 2023-11-28
Description: Automate monthly change for PL sandbox account passwords Issue: https://khanacademy.atlassian.net/browse/DIST-4092

rgagarinr:
Summary:
This is a cron job scheduled to run monthly Cypress/Python scripts for updating demo passwords in the PL Demo District.

Automate monthly change for PL sandbox account passwords.
Issue: DIST-4092

## Test plan:
Successfully ran the script manually on the Jenkins server.

### DIFF
--- a/jobs/demo-district-update-pass.groovy
+++ b/jobs/demo-district-update-pass.groovy
@@ -68,6 +68,7 @@ def runScript() {
    }
 }
 
+// We can't run this on the master because the Jenkins server doesn't have Cypress dependencies installed.
 onWorker("ka-test-ec2", '4d') {
    notify([slack: [channel: params.SLACK_CHANNEL,
                    sender: 'Testing Turtle',
--- a/jobs/demo-district-update-pass.groovy
+++ b/jobs/demo-district-update-pass.groovy
@@ -1,0 +1,85 @@
+// Automate monthly change for PL sandbox account passwords
+// https://khanacademy.atlassian.net/browse/DIST-4092
+
+@Library("kautils")
+// Classes we use, under jenkins-jobs/src/.
+import org.khanacademy.Setup;
+// Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
+//import vars.exec
+//import vars.notify
+//import vars.withTimeout
+//import vars.withSecrets
+
+new Setup(steps
+
+).allowConcurrentBuilds(
+
+).addStringParam(
+   "CYPRESS_GIT_REVISION",
+   """The name of a cypress branch to use when building.""",
+   "update-psw"
+
+).addStringParam(
+   "SLACK_CHANNEL",
+   "The slack channel to which to send failure alerts.",
+   "#demo-district-logs"
+
+).addStringParam(
+   "REVISION_DESCRIPTION",
+   """Set by the buildmaster to give a more human-readable description
+of the GIT_REVISION, especially if it is a commit rather than a branch.
+Defaults to GIT_REVISION.""",
+   "master"
+
+).addStringParam(
+   "TEST_RETRIES",
+   """How many retry attempts to use. By default is 1.""",
+   "1"
+
+).apply()
+
+// We use the build name as a unique identifier for user notifications. 
+BUILD_NAME = "build demo-district-password-update #${env.BUILD_NUMBER} (${params.REVISION_DESCRIPTION})"
+
+def _setupWebapp() {
+   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", params.CYPRESS_GIT_REVISION);
+   def runPythonTestArgs = 'sudo apt-get install -y python3.10'.split()
+   
+   dir("webapp/services/static") {
+      // We need to install Py3 to use google-api-spreadsheet-client
+      sh script: runPythonTestArgs.join(' '), returnStatus: true 
+      sh '/usr/bin/python3 -m pip install --upgrade google-api-python-client google-auth-httplib2 google-auth-oauthlib'
+   }
+}
+
+def runScript() {
+   def runCypressArgs = ["yarn",
+                         "cypress",
+                         "run", 
+                         "--spec",
+                         "javascript/districts-package/__e2e-tests__/change-pswd.spec.ts",
+                         "--browser=electron",
+                         "-c {\"retries\":${params.TEST_RETRIES}, \
+                         \"screenshotOnRunFailure\":false, \"video\":false}"
+   ];
+
+   dir('webapp/services/static') {        
+      exec(runCypressArgs); 
+      sh '/usr/bin/python3 /home/ubuntu/webapp-workspace/webapp/services/static/javascript/districts-package/__e2e-tests__/update-psw.py'
+   }
+}
+
+onWorker("ka-test-ec2", '4d') {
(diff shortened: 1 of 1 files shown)


jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,85 @@
...
+).addStringParam(
+   "SLACK_CHANNEL",
+   "The slack channel to which to send failure alerts.",
+   "#demo-district-logs"
+
+).addStringParam(
+   "REVISION_DESCRIPTION",
+   """Set by the buildmaster to give a more human-readable description
+of the GIT_REVISION, especially if it is a commit rather than a branch.
+Defaults to GIT_REVISION.""",
```
csilvers, line 31: Looks like this comment isn't true.  Should it have a default value (on line 32) at all?

jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+).addStringParam(
+   "TEST_RETRIES",
+   """How many retry attempts to use. By default is 1.""",
+   "1"
+
+).apply()
+
+// We use the build name as a unique identifier for user notifications. 
+BUILD_NAME = "build demo-district-password-update #${env.BUILD_NUMBER} (${params.REVISION_DESCRIPTION})"
```
csilvers, line 42: This is overkill for this job -- only one of these ever runs at a time.  You can look at other similar periodic jobs, like update-devserver-static-images.groovy, to see what kind of setup they do.

(I'd get rid of REVISION_DESCRIPTION entirely as well.)

csilvers (changes requested):
The overall logic looks good to me!  But some work is going to be required to set this up to run successfully in a way that doesn't stomp on other jobs.

jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,84 @@
...
+   """How many retry attempts to use. By default is 1.""",
+   "1"
+
+).addCronSchedule(
+   'H H 1 * *'        // Run on the first day of every month
+
+).apply()
+
+// We use the build name as a unique identifier for user notifications. 
+BUILD_NAME = "build demo-district-password-update #${env.BUILD_NUMBER}";
```
csilvers, line 38: I still don't think you need this -- the default BUILD_NAME is fine for you.

jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,85 @@
...
+                         \"screenshotOnRunFailure\":false, \"video\":false}"
+   ];
+
+   dir('webapp/services/static') {        
+      exec(runCypressArgs); 
+      sh '/usr/bin/python3 /home/ubuntu/webapp-workspace/webapp/services/static/javascript/districts-package/__e2e-tests__/update-psw.py'
+   }
+}
+
+onWorker("ka-test-ec2", '4d') {
```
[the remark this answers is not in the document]
csilvers: Still wondering about this.

jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,84 @@
...
+
+def _setupWebapp() {
+   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", params.CYPRESS_GIT_REVISION);
+   dir("webapp/services/static") {
+      sh("yarn install");
+   }
+}
+
+def runScript() {
+   def tempDir = "webapp/services/static/javascript/districts-package/__e2e-tests__";
```
csilvers, line 48: What makes this a "temp" dir?  This is a directory in our source tree, what's temporary about it?

jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,84 @@
...
+   def runCypressArgs = ["yarn",
+                         "cypress",
+                         "run", 
+                         "--spec",
+                         "javascript/districts-package/__e2e-tests__/change-pswd.spec.ts",
+                         "--browser=electron",
+                         "-c {\"retries\":${params.TEST_RETRIES}, \
+                         \"screenshotOnRunFailure\":false, \"video\":false}"
+   ];
+   def dockerBuild = ["docker", "build", "-t", "update-pass", "."];
```
csilvers, line 58: I don't see a Dockerfile in this directory.  Is that being added in another PR?

csilvers (commented):
Getting much closer!  But I'm not comfortable approving it until I understand more what's going on with the Dockerfile.

csilvers:
> @[rgagarinr](https://github.com/rgagarinr) rgagarinr requested a review from [csilvers](https://github.com/csilvers)

Just to be clear, the things I'm waiting on now:
1. Something pointing to this dockerfile that the job is using -- like the PR it's written in, or something
2. Some discussion of why you're running on a test-worker and not on master.

rgagarinr:
> > @[rgagarinr](https://github.com/rgagarinr) rgagarinr requested a review from [csilvers](https://github.com/csilvers)
> 
> Just to be clear, the things I'm waiting on now:
> 
> 1. Something pointing to this dockerfile that the job is using -- like the PR it's written in, or something
> 2. Some discussion of why you're running on a test-worker and not on master.

Sorry, I didn't submit my previous comments 🙇 they were in a pending state. Thanks again for reviewing @csilvers !
1) Here is a [dockerfile](https://github.com/Khan/webapp/commit/a1f8b7c12714b6887bcd32b4366e89fcf1c492a1#diff-ef08d52616c7afe9a5f5f4acc9a2e0f23813dc57cab71dde6e558198610ce634) from the db brach. 
2) Instead of on `master`? I tried it, but there is no `master`. What about `big-test-worker`? I'm okay with any worker that is safer and faster.

jobs/demo-district-update-pass.groovy
csilvers: Coolio.  I think a dockerfile is probably overkill here, but it should definitely work!

jobs/demo-district-update-pass.groovy
```diff
@@ -0,0 +1,84 @@
...
+      exec(runCypressArgs); 
+   }
+
+   dir(workDir) {
+      exec(dockerBuild); 
+      exec(dockerRun); 
+   }
+}
+
+onWorker("ka-test-ec2", '4d') {
```
csilvers, line 71:
```suggestion
onMaster('4d') {
```

csilvers (changes requested):
Change to onMaster and I think we're good to go!

jobs/demo-district-update-pass.groovy
rgagarinr: Thanks, I'll change it to OnMaster!

rgagarinr: @csilvers Seems like `OnMaster` is not [suitable](https://jenkins.khanacademy.org/job/misc/job/demo%20district/job/update-psw/41/console) for our (e2e) automation tasks, that's why we have been using the `ka-test-ec2` worker.

csilvers (approved):
Ah, good point.  OK, I'd add a comment to that effect ("we can't run this on master because jenkin-server doesn't have cypress deps installed"), and then you're good to go!

csilvers (approved):

PR #181, opened 2023-11-30
[🔥AUDIT🔥] Abort stuck webapp deploy jenkin's jobs

MiguelCastillo:
🖍 _This is an audit!_ 🖍

## Summary:
We have been monitoring webapp-deploy jenkins jobs that get stuck. This happens when someone in the front of the deploy queue aborts their deploy while jenkins is waiting for them to `set default`.  The issue seems to happen when we call `sh` to write to gcloud logs...

I had added logic in https://github.com/Khan/jenkins-jobs/pull/162 to tell us when the error happens by sending us a message in slack.  I had commented out the logic to throw an exception whenever we detect that the job was aborted while waiting for the `set default` prompt.  But now that I know the logic is correct, Im enabling the exception to abort the job if its stuck.

This should get us over the hump for now... I am also adding logging around the `sh` call to verify our theory.  If it turns out to be true, then the plan is to add a wrapper for `sh` that can check the status of the job when the call the `sh` is done.

## Test plan:
Deploy and wait for the slack message I added around the problematic `sh` call to tell us when a job is stuck.  If we get a confirmation hit, then I can add a `safeSh` type of function.

### DIFF
--- a/vars/notify.groovy
+++ b/vars/notify.groovy
@@ -388,11 +388,40 @@ def log(message, args=[:]) {
                           "--severity=${level} " +
                           logName + " " +
                           exec.shellEscape(jsonMessage));
+
+      def currentResult = currentBuild.result
+
       sh(shellCommand);
+
+      // TODO(miguel): remove this whole check once we confirm this is the
+      // the problem spot.
+      if (currentResult != currentBuild.result) {
+         // We have been dealing with aborted jobs getting stuck.  Traces point
+         // to `sh` calls not throwing exceptions when aborting jobs, so we
+         // want to verify that theory.  This particular case we are after is
+         // happening when we are in the webapp-deploy job, jenkin is waiting
+         // on the deployer to set default but instead the deployer aborts
+         // their deploy.  The theory is that `sh` should be throwing an
+         // an exception, but it never does in those situations.  So let's
+         // add extra logging to help us determine that.
+         def subject = "sh didn't abort ${env.JOB_NAME} ${env.BUILD_NUMBER}";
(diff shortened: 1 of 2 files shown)


vars/notify.groovy
```diff
@@ -388,11 +388,40 @@ def log(message, args=[:]) {
                           "--severity=${level} " +
                           logName + " " +
                           exec.shellEscape(jsonMessage));
+
+      def currentResult = currentBuild.result
+
       sh(shellCommand);
```
csilvers, line 394: `shellCommand` should totally be a list, and we should be using exec() here.  Not that it would help in itself, but then we could put this check in exec.groovy rather than here, and have it apply in more places.

csilvers (commented):
Sad we need this, but I'm glad we have it.  I really thought all these rethrowIfAborted's would do the trick!  Maybe one day we'll understand why they don't...

vars/notify.groovy
MiguelCastillo: Definitely!  The idea is to confirm that's what we need to do and then work my way thru all the spots the call `sh` and replace them with something that can detect aborted states like I added here.

I'd hate to have this show up again thinking we fixed it :(

MiguelCastillo:
> Sad we need this, but I'm glad we have it. I really thought all these rethrowIfAborted's would do the trick! Maybe one day we'll understand why they don't...

I know :(  This is just helping us avoid needing to babysit `#infrastructure-deploy`.  Im still going to look again at the `sh` command internal implementation when I have cycles.

PR #182, opened 2023-11-30
[🔥AUDIT🔥] Add more precise abort check

MiguelCastillo:
🖍 _This is an audit!_ 🖍

## Summary:
I added some logic in https://github.com/Khan/jenkins-jobs/pull/181 to send a message to slack if a job was aborted while in a call to `sh` and `sh` never throws an exception that aborts the jenkins job.  In this follow up change I am making the check more precise to ensure the job is in the `ABORTED` state.

## Test plan:
groovy vars/notify.groovy


csilvers (commented):
Ah, good idea

PR #185, opened 2023-12-07
Upload each image in a separate stage, for easier logs debugging.

csilvers:
## Summary:
This will also give us timing information on how long each stage takes.

I got rid of the `make fix_deps` because, by design, containers don't
require any local deps!  Any ones that are needed should built
explicitly by upload-images.sh first anyway, to make sure everything
is up to date.

I also made the merging of master optional.  It's something we'll
always want to do for the cronjob, but if you're running this in your
own branch you probably don't want it.

Issue: none

## Test plan:
I ran this manually at
https://jenkins.khanacademy.org/job/misc/job/update-devserver-static-images/123/

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -26,6 +26,17 @@ the updated data-file to it.""",
    "automated-commits"
 
 ).addStringParam(
+    "SERVICES",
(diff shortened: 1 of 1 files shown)


dbraley (approved):
Nice!

PR #186, opened 2023-12-09
Retry upload-images.sh on failure.

csilvers:
## Summary:
The update-devserver-static-images job periodically (regularly,
but unpredictably) fails with:
> ERROR: failed to receive status: rpc error: code = Unavailable desc = error reading from server: EOF

I have been unable to figure out why.  Until that's figured out, let's
just retry.  I'm hoping it's rare enough we can get through this,
eventually, if we retry enough.

Issue: https://jenkins.khanacademy.org/job/misc/job/update-devserver-static-images/130/execution/node/196/log/

## Test plan:
Will run it on jenkins


MiguelCastillo (commented):
👍

PR #188, opened 2023-12-19
Remove all references to deploying to GAE from the jenkins jobs.

csilvers:
## Summary:
Ever since https://github.com/Khan/webapp/pull/11178, Jenkins has not
deployed the dynamic (python2) service unless specifically asked to do
so via something like `sun: deploy mybranch (SERVICES=dynamic)`.  But
the code to deploy to default still existed.

This PR rips it out.  Way out!  Because the SAT service has been
sunset (sunsetted?), and there is no code running on these services
anymore.  If you ask to deploy the `dynamic` service you'll just get
an error.

This should wait until after 1 Jan 2024 to land, since that's when
the SAT product is formally sunset(ted).

Issue: https://khanacademy.atlassian.net/browse/INFRA-9701

## Test plan:

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -381,7 +381,9 @@ def deployToGCS() {
    def args = ["deploy/deploy_to_gcs.py", NEW_VERSION,
                "--slack-channel=${params.SLACK_CHANNEL}",
                "--deployer-username=${DEPLOYER_USERNAME}",
-               // Same as for deploy_to_gae, we suppress the changelog for now.
+               // We don't send the changelog in a build-only context:
+               // there may be many builds afoot and it is too
+               // confusing.  We'll send it in promote instead.
                "--suppress-changelog",
                // Since we're deploying new static code, we should upload the
                // updated sourcemap files to our error reporting system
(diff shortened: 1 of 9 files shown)


nathanjd (approved):

jobs/build-webapp.groovy
```diff
@@ -76,11 +77,10 @@ new Setup(steps
     """<p>A comma-separated list of services we wish to deploy (see below for
 options), or the special value "auto", which says to choose the services to
 deploy automatically based on what files have changed.  For example, you might
-specify "dynamic,static" to force a full deploy to GAE and GCS.</p>
+specify "users,static" to force a full deploy to the users service and GCS.</p>
```
nathanjd, line 80: I think we'll want to make the same changes to the deploy-znd services parameter as well.

jobs/deploy-znd.groovy
```diff
@@ -175,57 +174,18 @@ def determineVersion() {
    }
 }
 
-
-// This should be called from within a node().
-def deployToGAE() {
```
MiguelCastillo, line 180: YAY!!!

MiguelCastillo (approved):
Excellent!

PR #191, opened 2024-01-10
Temporarily disable first smoke tests

jandrade:
## Summary:

This is a temporary measure while we investigate the cause of the slowness in
our e2e testing infra.

Issue: XXX-XXXX

## Test plan:

Verify that this replay worked as expected:

- First smoke test: https://jenkins.khanacademy.org/job/deploy/job/e2e-test/55630/console

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -312,27 +312,30 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
            buildmaster: [sha: params.GIT_REVISION,
                          what: E2E_RUN_TYPE]]) {
 
-      // NOTE(juan): This is a temporary hack to notify the buildmaster that
-      // we are done with the first smoke test. This is needed because of some
-      // issues with our E2E testing infrastructure.
       if (E2E_RUN_TYPE == "first-smoke-test") {
-         // send signal to buildmaster.
-         echo("Notifying buildmaster that we are done with first smoke test");
-      } else {
-         stage("Sync webapp") {
-            _setupWebapp();
-         }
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -300,31 +300,40 @@ def analyzeResults() {
...
-         stage("Run e2e tests") {
-            runLambdaTest();
+                         what: E2E_RUN_TYPE]]) {
+
+      // NOTE(juan): This is a temporary hack to notify the buildmaster that
+      // we are done with the first smoke test. This is needed because of some
+      // issues with our E2E testing infrastructure.
+      if (E2E_RUN_TYPE == "first-smoke-test") {
+         // send signal to buildmaster.
+         echo("Notifying buildmaster that we are done with first smoke test");
```
csilvers, line 320: This is a bit misleading.  I don't think you need to talk about notifying buildmaster here; I think the important thing here is that you're just running a faster (much faster) command.

You can also use `return` here so you don't have to indent.  Something like:
```
if (E2E_RUN_TYPE == "first-smoke-test") {
       // NOTE(juan): This is a temporary hack to just not run first
       // smoke tests.  This is needed because some issues with
       // our E2E testing infrastructure are causing smoketests to
       // take a very long time to run, so we can't afford to run two
       // of them for every deploy.  We skip the first smoke test
       // because it takes the longer of the two (since pre-default
       // webapp isn't as well primed).
       echo("Temporarily skipping first smoke tests");
       return;
}
```

csilvers (approved):
Looks good!  A small suggestion for better documentation inline.

PR #192, opened 2024-01-11
Revert change that disabled first smoke tests

jandrade:
## Summary:

After disabling the first smoke tests (#191), we've seen that the LambdaTest
infra is more stable. This gives us confidence that the first smoke tests are
back to how they were working before.

This PR reverts the change that skipped the first smoke tests and instead runs
the regular flow in the e2e-tests job.

Issue: XXX-XXXX

## Test plan:

N/A

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -312,17 +312,6 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
            buildmaster: [sha: params.GIT_REVISION,
                          what: E2E_RUN_TYPE]]) {
 
-      if (E2E_RUN_TYPE == "first-smoke-test") {
-         // NOTE(juan): This is a temporary hack to just not run first smoke
-         // tests.  This is needed because some issues with our E2E testing
-         // infrastructure are causing smoketests to take a very long time to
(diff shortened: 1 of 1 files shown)


somewhatabstract (approved):
This looks good to me.

I wonder if we should consider building in some kind of switch that uses the "Skip tests" sheet to signal when we want this so we don't have to do a code change should we need to do this again.

jandrade:
> This looks good to me.
> 
> I wonder if we should consider building in some kind of switch that uses the "Skip tests" sheet to signal when we want this so we don't have to do a code change should we need to do this again.

That's a good observation. I'm going to file a ticket to see if/how we could approach that.

MiguelCastillo:
> > This looks good to me.
> > I wonder if we should consider building in some kind of switch that uses the "Skip tests" sheet to signal when we want this so we don't have to do a code change should we need to do this again.
> 
> That's a good observation. I'm going to file a ticket to see if/how we could approach that.

I was thinking about adding just this yesterday.  We can probably pass the parameter `SKIP_TESTS` and fill it out from buildmaster with the correct value.  Perhaps `SKIP_TESTS="*"` or something.

PR #195, opened 2024-01-24
Improve e2e instrumentation

jandrade:
## Summary:

This PR improves the e2e instrumentation by slightly changing the way we build
our e2e tests:

1. Adding tags to each spec, so we can easily filter them out in the future via
the Dashboard UI or the API.
2. Notifying to the #fei-testing-logs slack channel in case of "first/second
smoke test" failures.

Issue: XXX-XXXX

## Test plan:

Verify that the e2e tests are still running as expected and that the tags are
included in the test results.

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -183,6 +183,7 @@ def runLambdaTest() {
                             "-p=${params.NUM_WORKER_MACHINES}",
                             "--sync=true",
                             "--bt='jenkins,${e2eEnv}'",
+                            "--tags='jenkins,${e2eEnv}'",
                             "--eof"
    ];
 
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -183,6 +183,7 @@ def runLambdaTest() {
                             "-p=${params.NUM_WORKER_MACHINES}",
                             "--sync=true",
                             "--bt='jenkins,${e2eEnv}'",
+                            "--tags='jenkins,${e2eEnv}'",
```
jandrade, line 186: **note:** We have to duplicate the tags here (bt and tags) due to the way LambdaTest exposes its CLI :|
- `bt` stands for `buld-tags`, which are tags associated to a build (or a session that runs multiple test specs).
- `tags` are the tags associated to each test spec.

csilvers (approved):

jeresig (commented):
Nice - thanks for improving this!

jobs/e2e-test.groovy
MiguelCastillo: Thanks for this comment.  It would be very useful to include it here since to the untrained eye the duplication does not seem obviously intentional.  For the future tho.

PR #196, opened 2024-01-25
Add support for enabling the python3 virtualenv when running jobs.

csilvers:
## Summary:
We are porting our deploy and testing scripts from python2 to python3.
Some of them need third-party libraries such as `requests` or
`graphql`, which we are supporting via a python3 virtualenv that is
constructed in webapp's deploy/Makefile.

This PR adds Jenkins support to use that virtualenv.  We don't have a
user-friendly way to enable the virtualenv -- we don't expect users to
run these scripts, just Jenkins -- but the new
`withVirtualenv.python3()` does all the needful.

I had a choice: I could set up this new virtualenv in onWorker() and
onMaster(), like we do with the python2 virtualenv, or I could require
jobs to call it manually to set it up in the places where it's needed.
I chose to do the latter, since our goal is to replace all our python3
with Go (or js) over time, and I'm hoping that will allow us to remove
these calls as that happens, and eventually be left with no
withVirtualenv calls at all.  That is easier to do one at a time than
when it's globally done in onMaster/onWorker.

It does make it harder to know when to call the new python3() call,
since we need to make sure we're on the right machine when we do it.
For now, that's only an issue with webapp-test.groovy, which runs
across a lot of machines.

Issue: https://khanacademy.atlassian.net/browse/INFRA-9701

Test plan:

I ran `groovy` on all the files in this commit, with only the normal,
expected errors around `@CPS` and the like.

### DIFF
--- a/vars/withVirtualenv.groovy
+++ b/vars/withVirtualenv.groovy
@@ -12,7 +12,7 @@ def call(Closure body) {
    if (env.VIRTUAL_ENV && fileExists(env.VIRTUAL_ENV)) {
       // we're already in a virtualenv.  The fileExists is necessary
       // because a node can inherit the environment from its parent
-      // and have an inaccuarte VIRTUAL_ENV.
+      // and have an inaccurate VIRTUAL_ENV.
       echo("(Not activating virtualenv; already active at ${env.VIRTUAL_ENV})");
       // Make sure the virtual-env is still first in the path.
       withEnv(["PATH=${env.VIRTUAL_ENV}/bin:${env.PATH}"]) {
@@ -60,6 +60,18 @@ For more information, see https://wiki.python.org/moin/DebuggingWithGdb
    withEnv(["VIRTUAL_ENV=${pwd()}/env",
             "PATH=${pwd()}/env/bin:${env.PATH}"]) {
       _maybeDowngradePip();
+      body();
+   }
+}
+
+// NOTE: it is safe to have both the above and below active at the same
+// time.  `python2` binaries will use the virtualenv from call(), and
+// `python3` binaries will use the virtual from here.
+// This must be called from workspace-root.
+def python3(Closure body) {
+   echo("Activating python3 virtualenv");
+   sh("make -C webapp/deploy deps");
+   withEnv(["PATH=${pwd()}/webapp/deploy/venv/bin:${env.PATH}"]) {
       body();
    }
 }
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -40,6 +40,7 @@ import org.khanacademy.Setup;
 //import vars.onWorker
 //import vars.withSecrets
 //import vars.withTimeout
+//import vars.withVirtualenv
 
 
 def formatServicesList(services) {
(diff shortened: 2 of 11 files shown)


vars/withVirtualenv.groovy
```diff
@@ -63,3 +63,15 @@ For more information, see https://wiki.python.org/moin/DebuggingWithGdb
...
 }
+
+// NOTE: it is safe to have both the above and below active at the same
+// time.  `python2` binaries will use the virtualenv from call(), and
+// `python3` binaries will use the virtual from here.
+// This must be called from workspace-root.
+def python3(Closure body) {
+   echo("Activating python3 virtualenv");
+   sh("make -C webapp/deploy deps");
+   withEnv(["PATH=${pwd()}/webapp/deploy/venv/bin:${env.PATH}"]) {
```
MiguelCastillo, line 74: Neat!  I like this.

vars/withVirtualenv.groovy
```diff
@@ -63,3 +63,15 @@ For more information, see https://wiki.python.org/moin/DebuggingWithGdb
       body();
    }
 }
+
+// NOTE: it is safe to have both the above and below active at the same
+// time.  `python2` binaries will use the virtualenv from call(), and
+// `python3` binaries will use the virtual from here.
+// This must be called from workspace-root.
```
MiguelCastillo, line 70: Is it worth adding a TODO to remove when we no longer use python3?

MiguelCastillo (approved):
I like the solution!  It gets us over the hump quite nicely.

Thanks for calling out the drawbacks too.

> It does make it harder to know when to call the new python3() call,

This is the tricky thing we have been working with since we've had p2 and p3.  So happy you are making tons of progress on getting rid of p2.

> since we need to make sure we're on the right machine when we do it.
For now, that's only an issue with webapp-test.groovy, which runs
across a lot of machines.

This one is trickier to me.  Im not sure I understand what "the machine" means.  Are we not installing python3 in all the workers and the server?

csilvers:
> This one is trickier to me. Im not sure I understand what "the machine" means. Are we not installing python3 in all the workers and the server?

Consider this:
```
withVirtualenv.python3() {
   deploy/foo.py  # great
   onWorker() {
       deploy/bar.py   # not great, it's on a separate machine and does not see the virtualenv
   }
}
```
This kind of thing actually happens in webapp-test.groovy.

vars/withVirtualenv.groovy
csilvers: I don't think it's necessary, though it may be beneficial to have TODOs in each of the places we call withVirtualenv.python3().  But that's a lot of places!

MiguelCastillo:
> > This one is trickier to me. Im not sure I understand what "the machine" means. Are we not installing python3 in all the workers and the server?
> 
> Consider this:
> 
> ```
> withVirtualenv.python3() {
>    deploy/foo.py  # great
>    onWorker() {
>        deploy/bar.py   # not great, it's on a separate machine and does not see the virtualenv
>    }
> }
> ```
> 
> This kind of thing actually happens in webapp-test.groovy.

ah - that's very helpful. thanks for the illustration.

PR #198, opened 2024-02-02
Remove args that just don't get used

somewhatabstract:
## Summary:
The removed args don't do anything - they aren't passed along to the CLI invocation because we manage that stuff internally.

These are the args we support: https://github.com/Khan/webapp/blob/04b6b63ae70d1969427140601b0961e9540cb93f/services/static/dev/cypress/e2e/util/parse-lambdatest-arguments.ts#L34-L42

Related change for webapp GitHub actions: https://github.com/Khan/webapp/pull/19586

Issue: FEI-5354

## Test plan:
There's nothing to test really - these args were already noops disappearing into the ether. However, on putting this change into rotation, we will see that all still works and nothing changed.


kevinbarabash (approved):
Thanks for updating the jenkins job! 🚀

PR #206, opened 2024-02-07
[🔥AUDIT🔥] Reverting forcing make deps in deploy scripts

MiguelCastillo:
🖍 _This is an audit!_ 🖍

## Summary:
I forced the creation of the virtual env for python in the jobs to help me determine if that would make python happy creating the virtual env.  It does. But clearly that is not what we want to do.  So Im reverting while I figure out a sustainable solution.

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1707318372614619

## Test plan:
Should be back to what we were before.  Im just undoing the change.


csilvers (commented):
Is it safe to revert this -- won't it cause builds to fail again?  Anyway, I'm fine with this if you think it's safe!

PR #207, opened 2024-02-07
When downgrading the pip, make sure we're downgrading the one in the virtualenv.

csilvers:
## Summary:
We're seeing pip at version 19 in the deploy (python3) virtualenv,
which shouldn't be happening!  We have an (unproven) theory that the
issue is that somehow our pip-downgrading is affecting that virtualenv
rather than the python2 virtualenv.  This PR should make that
impossible.

Issue: https://jenkins.khanacademy.org/job/deploy/job/build-webapp/lastFailedBuild/console

## Test plan:
Fingers way crossed


MiguelCastillo (approved):
let's give it a try!

PR #210, opened 2024-02-12
Do a better job of indicating task failures in the weekly-maintenance script.

csilvers:
## Summary:
This change makes it so each failing task has a red dot next to it on
the pipeline-steps page.  Before this, they all had green dots, and
you had to look at the message at the end of the job to figure out
which tasks had failed.

(Now there is no message at the end of the job, which is maybe a
regression in quality but I think the red dots will be more useful.)

Issue: none

## Test plan:
I manually re-ran this new code at
https://jenkins.khanacademy.org/job/misc/job/webapp-maintenance/293/console


MiguelCastillo (approved):
Oh yeah this is a nice change! thank you.

PR #211, opened 2024-02-14
[🔥AUDIT🔥] Make sure we run our fastly deploys with the python3 virtualenv.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
It looks like our deploy_fastly.py script needs to import third-party
libs, which means we need to use the virtualenv!  I just missed this
before because the jenkins jobs doesn't call a python script directly,
it calls a make rule that calls a python script.

Issue: https://khanacademy.slack.com/archives/CCCC6HZ4K/p1707924185156879

## Test plan:
groovy jobs/deploy-fastly.groovy

Subscribers: @jandrade, @jeresig

### DIFF
--- a/jobs/deploy-fastly.groovy
+++ b/jobs/deploy-fastly.groovy
@@ -26,6 +26,7 @@ import org.khanacademy.Setup;
 //import vars.notify
 //import vars.withSecrets
 //import vars.withTimeout
+//import vars.withVirtualenv
 
 
 new Setup(steps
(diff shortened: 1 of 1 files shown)


MiguelCastillo:
Good catch!  It so easy to miss this stuff.  And even more subtle is changing something in webapp and encountering these errors that don't seem related and are kinda difficult to correlate.

PR #214, opened 2024-02-16
No longer run `make python_deps`.

csilvers:
## Summary:
This builds the python2 deps, but we no longer need that, because we
no longer run any python2 code from jenkins!

Issue: https://khanacademy.atlassian.net/browse/INFRA-9771

## Test plan:
None


MiguelCastillo (approved):
Great!

PR #216, opened 2024-02-21
Don't verify first-smoke-test if only static service changed.

jeresig:
## Summary:
This change is made together with:
https://github.com/Khan/buildmaster2/pull/170

We don't need to verify the first-smoke-test when only the static service is being deployed, as no e2e tests are running in that case!

Issue: FEI-5450

## Test plan:
Unsure!

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -564,10 +564,10 @@ def _maybeAbortJob(oldBuildResult, newBuildResult, reason) {
       // to keep track of these cases.
       _simpleSlackAlert(
          "#infrastructure-deploys",
-         """Deploy aborted but thread is still running. We most likely have a 
-job that is stuck. ${reason}. prev state ${oldBuildResult}. current state 
-${newBuildResult}. ${env.BUILD_URL}. 
-See https://khanacademy.atlassian.net/wiki/spaces/INFRA/pages/2470543393/Stuck+jenkins+deploy-webapp+builds 
+         """Deploy aborted but thread is still running. We most likely have a
+job that is stuck. ${reason}. prev state ${oldBuildResult}. current state
+${newBuildResult}. ${env.BUILD_URL}.
+See https://khanacademy.atlassian.net/wiki/spaces/INFRA/pages/2470543393/Stuck+jenkins+deploy-webapp+builds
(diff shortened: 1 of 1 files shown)


kevinbarabash (approved):
LGTM 🚢

jobs/deploy-webapp.groovy
```diff
@@ -955,7 +955,10 @@ onMaster('4h') {
 
       try {
          stage("Await first smoke test") {
-            if (SERVICES) {
+            // Only verify smoke tests if we're deploying a non-static service
+            // Static service changes have already had e2e tests run on them
+            // during the PR process, so this is unnecessary.
```
csilvers, line 960: This is an exaggeration -- these tests run against the commit + master (and possibly + the prior commit), which is exactly the kind of situation that we're worried about when we run these smoke tests in the first place: that a bad merge or something will cause problems we couldn't detect by just testing the commit itself in isolation.

csilvers (changes requested):
I think it's reasonable to consider turning off first e2e tests in this situation, though it comes with a cost, as mentioned below.  But I think if we do it should be in the context of ADR 699 -- we should probably augment it to talk about this case?

It seems to me, we could make the same arguments here as for running tests before deploy: we run tests against every PR as well.  And in neither case is a failure blocking landing the PR.  (Another thing I'm worried about if we land this is that someone will have a static-only PR that does go through the github-action smoketest, which will report a failure, which the author doesn't see, perhaps because the PR was approved and landed so quickly the smoketest never had time to run.)

Overall I'd like to discuss this case more before landing it.  The backend guild meeting on Friday would be a perfect time.  I will do my best to make it.

jobs/deploy-webapp.groovy
csilvers, line 960:
```suggestion
            // during the PR process, so doing it here adds little value.  (Its only job would
            // be to catch bad merges between this deploy and other deploys between the
            // time this PR was created and now.)
```

csilvers (approved):
OK, based on the discussion at the backend guild meeting today I'm approving!  I also updated https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/2643591176/ADR+699+Allow+for+less-attended+and+unattended+deploys to include this as part of the Decision.

PR #219, opened 2024-02-29
Turn off caniuse weekly upgrade, for now.

jeresig:
## Summary:
This is going to run and that will cause the queryplanner and render-gateway services to attempt to deploy. However, the GCP node16 runtime is deprecated and those services can no longer be deployed. So we need to turn this off until we can successfully deploy this services again, hopefully next week.

Issue: FEI-5536

## Test plan:
The weekly maintenance job should run without errors, I hope!

### DIFF
--- a/weekly-maintenance-jobs.sh
+++ b/weekly-maintenance-jobs.sh
@@ -285,26 +285,28 @@ clean_package_files() {
     jenkins-jobs/safe_git.sh commit_and_push webapp -a -m "Automatic cleanup of language package files"
 }
 
-
-update_caniuse() {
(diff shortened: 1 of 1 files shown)


somewhatabstract (approved):
LGTM. Thank you.

csilvers:
Now that we've deployed the node20 change, can we turn this back on again?

jeresig:
If it's been deployed, then yes! I'll just revert this commit.

jeresig:
I just pushed the revert, so we should be good now. Thanks for the heads-up!

PR #221, opened 2024-04-05
[🔥AUDIT🔥] Allow should_deploy.py to access slack when *deploying* to webapp.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
Slack is temporarily needed while should_deploy is doing side-by-side
testing of the go_code_analysis.go.

I only do this in deploy-webapp.groovy, though should_deploy.py is
also called in build_webapp.groovy, so we only get notified once per
deploy, not lots of times (as a deploy moves up the queue).  This not
only avoids spam, it makes it easier to figure out which deploy a
slack message corresponds to, since only one deploy-webapp job can be
running at a time.  Since both jobs give the same output, this is safe
to do.

Issue: none

## Test plan:

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -400,15 +400,21 @@ def mergeFromMasterAndInitializeGlobals() {
 
             // TODO(benkraft): Extract the resolution of services into a util
             if (params.SERVICES == "auto") {
-               try {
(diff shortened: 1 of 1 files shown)


mpolyak:
> only one deploy-webapp job can be running at a time.

Nice, that should help parsing the messages.

MiguelCastillo (commented):
👍

PR #222, opened 2024-04-11
Removed debug slack message when deploys get stuck in jenkins

MiguelCastillo:
## Summary:
We had a nasty issue sometime ago in this PR https://github.com/Khan/jenkins-jobs/issues/181.  I have since been able to confirm that the logic is correctly aborting the jobs so I am removing the messages we are sending to slack about stuck jobs.

Issue: https://khanacademy.slack.com/archives/C06385ANLAC/p1712855669448049

## Test plan:
deploy and never see the message again :)


csilvers (approved):
Good job, Mr. Anderson.

PR #223, opened 2024-07-08
Don't have a gc failure be a job failure.

csilvers:
## Summary:
Sometimes when gc-ing in a workspace, we get an error like this:
```
error: Could not read 1f64cbb3aab35f38a0ba7bf53ea29ae399821dd3
fatal: Failed to traverse parents of commit d2125172e642444012fc1199df32302195d98eec
fatal: failed to run repack
```

I have no idea why -- running the gc manually works fine -- but it's
not a big deal if we can't gc everything all the time, so we just let
it be.  This should quiet the error messages on the weekly maintenance
job, hopefully.

Issue: https://jenkins.khanacademy.org/job/misc/job/webapp-maintenance/316/execution/node/239/log/

## Test plan:
Fingers crossed

### DIFF
--- a/weekly-maintenance-jobs.sh
+++ b/weekly-maintenance-jobs.sh
@@ -102,7 +102,14 @@ gc_all_repos() {
         echo "GC-ing in $dir"
         cd "$dir"
 
-        git gc
+        # Sometimes we get an error like this:
+        #   error: Could not read 1f64cbb3aab35f38a0ba7bf53ea29ae399821dd3
+        #   fatal: Failed to traverse parents of commit d2125172e642444012fc1199df32302195d98eec
+        #   fatal: failed to run repack
+        # I have no idea why -- running the gc manually works fine --
+        # but it's not a big deal if we can't gc everything all the
+        # time, so we just let it be.
+        gc || echo "Failed to GC in $dir, continuing anyway"
         )
     done
 }


MiguelCastillo (approved):
Seems reasonable

weekly-maintenance-jobs.sh
```diff
@@ -102,7 +102,14 @@ gc_all_repos() {
...
 
-        git gc
+        # Sometimes we get an error like this:
+        #   error: Could not read 1f64cbb3aab35f38a0ba7bf53ea29ae399821dd3
+        #   fatal: Failed to traverse parents of commit d2125172e642444012fc1199df32302195d98eec
+        #   fatal: failed to run repack
+        # I have no idea why -- running the gc manually works fine --
+        # but it's not a big deal if we can't gc everything all the
+        # time, so we just let it be.
+        gc || echo "Failed to GC in $dir, continuing anyway"
```
MiguelCastillo, line 112: Ok - I looked a bit closer.  I think the part of the message that is most helpful is `warning: reflog of 'HEAD' references pruned commits`, which we are not including in here.  I think that's the actual cause of the failure.

Can we try running `git reflog expire --all --stale-fix` before running `gc`?

We don't do anything with the reflogs in jenkins, so it should be fine to aggressively prune.  Or perhaps we add that in the merge-branches job itself.  We see this when we run `git fetch --tags --prune --prune-tags --force --progress origin`.  See https://jenkins.khanacademy.org/job/deploy/job/merge-branches/250041/console

Check out this post about this same issue here https://feeding.cloud.geek.nz/posts/error-while-running-git-gc/

csilvers: Would love to give that a try!  I agree that doing it all the time, rather than just in the weekly-cleanup, makes sense.  It sounds like I should do it before any fetch/pull?

MiguelCastillo: Yup I think that's a good spot to do it.

csilvers:
#224 seemed to fix the underlying cause, so this doesn't seem needed anymore!

PR #224, opened 2024-07-09
Clean the reflog before pulling or gc-ing.

csilvers:
## Summary:
Every so often -- very commonly when running the weekly maintenence gc
scripts, but at other times like `merge-branches` as well -- we see an
error like:
> warning: reflog of 'HEAD' references pruned commits

@Miguelcastillo did some sleuthing and found the fix, which I put
here.  Since every `pull` can cause a `gc`, I just clean the reflog
before every pull; we don't need the reflog in Jenkins, since it's
really a tool for manual git work and Jenkins is fully automated.

Issue: https://jenkins.khanacademy.org/job/misc/job/webapp-maintenance/316/execution/node/239/log/

## Test plan:
I ran the new `git reflog` command manually on jenkins in
~jenkins/jobs/misc/jobs/webapp-maintenance/workspace/webapp,
and it ran ok.

### DIFF
--- a/safe_git.sh
+++ b/safe_git.sh
@@ -83,6 +83,10 @@ _alert() {
 
 # Call this from within the repo that you want to do the fetching.
 _fetch() {
+    # We don't need reflogs in jenkins, and they can cause trouble
+    # when gc-ing, so we remove them.  See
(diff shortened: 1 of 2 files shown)


MiguelCastillo (approved):
Awesome! Crossing fingers this makes the issue go away 🤞

PR #225, opened 2024-07-15
Use official update package/script for browserslist package update

jeremywiebe:
## Summary:

EDIT: I've updated this PR to use the official script (`update-browserslist-db`) instead of the manual `sed` + `yarn` system.

https://github.com/browserslist/update-db

~Our `weekly-maintenance-jobs.sh` script updates the browserslist and caniuse-lite packages. In the `yarn` invocation, we're using the wrong package name. Thankfully, there is a typo package (`browserlist`) that points to the right package. But long term we should be using the correct name (`browserslist` (note the `s` suffix)).~

~See: https://www.npmjs.com/package/browserlist which reads:~

~> Package to help in popular typo: browserlist instead of browserslist~

image

Issue: "none"

## Test plan:

:fingerscrossed:

### DIFF
--- a/weekly-maintenance-jobs.sh
+++ b/weekly-maintenance-jobs.sh
@@ -296,10 +296,8 @@ update_caniuse() {
         for d in `git grep -l caniuse-lite "*yarn.lock" | xargs -n1 dirname`; do
             (
                 cd "$d"
-                # This deletes everything from the first "caniuse-lite" line
-                # to the following blank line, from yarn.lock.
-                sed -i '/^caniuse-lite@/,/^$/d' yarn.lock
-                yarn upgrade caniuse-lite browserslist
+                # Use the official tool to update the browserslist and caniuse-lite packages.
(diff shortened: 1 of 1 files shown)


jeresig (approved):
lol, thank you!

I wonder if we can replace all of this logic and just do this instead:

```
npx update-browserslist-db@latest
```

jeremywiebe:
> lol, thank you!
> 
> I wonder if we can replace all of this logic and just do this instead:
> 
> ```
> npx update-browserslist-db@latest
> ```

Potentially! Are we ok, from a security point of view, running `npx` in automation?

jeresig:
@jeremywiebe Yeah, I think that's fine! Especially since this is the official tool that they recommend running.

jeremywiebe:
I'll update this PR!

jeresig (approved):
Great - thank you Jeremy!

PR #226, opened 2024-07-17
[🔥AUDIT🔥] Do not do reflog cleaning all the time, only in the weekly maintenance.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
It's slow, and we believe it's causing errors (perhaps when the reflog
and gc coincide).

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1721248190898759

## Test plan:
Fingers crossed

Subscribers: @somewhatabstract

### DIFF
--- a/safe_git.sh
+++ b/safe_git.sh
@@ -83,10 +83,6 @@ _alert() {
 
 # Call this from within the repo that you want to do the fetching.
 _fetch() {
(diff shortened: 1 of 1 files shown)


MiguelCastillo:
👍

PR #228, opened 2024-08-27
[🔥AUDIT🔥] Run more ownership_data tests when updating ownership data.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
When we update the ownership data in our nightly deploy, we run some
tests to make sure that the new ownership data doesn't break anything.
It turns out that some potential problems are tested in a _different_
ownership test, so we need to run that one too.

Issue: https://khanacademy.slack.com/archives/C02NMB1R5/p1724751059899299

## Test plan:
I ran the new command manually on jenkins

Subscribers: @jeresig


jeresig:
Great, thank you Craig!

PR #229, opened 2024-08-28
Refactor the jenkins jobs so they don't run `clean` after building deps.

csilvers:
## Summary:
Several of our jobs run `virtualenv.withPython3` at a very high level
(in main).  And then call a function that calls `clean()`.
Unfortunately, that's not safe: `withPython3` builds python deps that
`clean()` can nuke.  Luckily, this doesn't happen by default: you have
to run with `CLEAN=all`.  But it's still a problem.

This PR refactors the jobs so they always run `withPython3` after
`clean`, rather than before.  A better fix would be to only run
`withPython3` in the places we're actually running python3 code, not
at this high level in `main()`, but that's a more significant
refactor, for another day.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1724860397897649

## Test plan:
Mostly fingers crossed -- I'll try the jobs on Jenkins.  But I did run:
```
groovy jobs/deploy-znd.groovy
```
which showed no syntax errors.

Subscribers: @somewhatabstract

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -381,8 +381,6 @@ def mergeFromMaster() {
 def initializeGlobals() {
    withTimeout('1h') {    // should_deploy builds files, which can take forever
       dir("webapp") {
-         clean(params.CLEAN);
-
          // Let's do a sanity check.
          def headSHA1 = exec.outputOf(["git", "rev-parse", "HEAD"]);
          if (params.GIT_REVISION != headSHA1) {
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -324,8 +324,8 @@ def _sendAbortedDeploymentSurvey() {
     }
 }
 
-def mergeFromMasterAndInitializeGlobals() {
-   withTimeout('1h') {    // should_deploy builds files, which can take forever
+def mergeFromMaster() {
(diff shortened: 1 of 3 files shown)


jobs/deploy-webapp.groovy
```diff
@@ -372,79 +372,85 @@ def mergeFromMasterAndInitializeGlobals() {
...
+   }
+}
 
-            // Let's do a sanity check.
-            def headSHA1 = exec.outputOf(["git", "rev-parse", "HEAD"]);
-            if (params.GIT_REVISION != headSHA1) {
-               notify.fail("Internal error: " +
-                           "HEAD does not point to the deploy-branch");
-            }
+def initializeGlobals() {
```
csilvers, line 381: This will be easier to read if you tell git to ignore whitespace (via the gear).

jobs/deploy-webapp.groovy
```diff
@@ -916,7 +922,12 @@ onMaster('4h') {
                          what: 'deploy-webapp']]) {
       try {
          stage("Merging in master") {
-            mergeFromMasterAndInitializeGlobals();
+            mergeFromMaster();
+         }
+         stage("Initializing globals") {
+            withVirtualenv.python3() {
+               initializeGlobals();
```
nathanjd, line 929: Looks like this one still has a clean happening on line 384?

PR #232, opened 2024-09-26
Fix the deploy-fastly script.

csilvers:
## Summary:
I landed it with some uncalled functions and repeated functionality.
I also landed it when we didn't have a "set-default" option for
compute@edge.  But now we do, so let's fix up the jenkins job.  It's a
lot cleaner now.

Issue: none

## Test plan:
Fingers crossed -- will try to run it once it's landed!


MiguelCastillo (approved):
This makes a lot more sense! thank you.

PR #233, opened 2024-09-27
Deploy to fastly-staging as part of the deploy process.

csilvers:
## Summary:
The idea is that if you are deploying fastly as part of your deploy,
we want the pre-set-default url
(`https://prod-2409<etc>.khanacademy.org`), to execute the new fastly
code.  The way we do this is to deploy the new fastly code to a
separate "staging" fastly service, and then route all requests to
prod-etc.ka.org to that staging fastly service instead of the regular
prod fastly service.

The details of how we do that routing are beyond the scope of this PR,
but this PR _does_ do the important step of making sure that the
"staging" fastly service has the code for this deploy in it!  And it
makes sure that happens before we do the "set default" prompt, so that
any manual testing that is done gets the new fastly.

Sadly, the first-smoke-test run does _not_ hit the new fastly, it
still hits the old one.  This is unavoidable because we can run
multiple fisrt-smoke-test runs at the same time (one for every deploy
in the deploy queue), but we only have one "staging" fastly service,
which we reserve for the deploy at the front of the deploy queue.
I've added a NOTE about that, we will have to try to figure out some
way to handle this, later.

Issue: https://khanacademy.atlassian.net/browse/INFRA-10385

## Test plan:

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -478,6 +478,20 @@ def _manualSmokeTestCheck(job){
 }
 
 
+def deployFastlyStaging() {
+   withTimeout('10m') {
+      withSecrets.slackAndStackdriverAlertlibOnly() {
+         exec(["make", "-C", "webapp/services/fastly-khanacademy-compute",
+               "deploy-staging",
+               "ALREADY_RAN_TESTS=1",
+               "DEPLOY_VERSION=${NEW_VERSION}"]);
+         exec(["make", "-C", "webapp/services/fastly-khanacademy-compute",
+               "set-default-staging",
+               "DEPLOY_VERSION=${NEW_VERSION}"]);
+      }
+   }
+}
+
 def verifySmokeTestResults(jobName, buildmasterFailures=0) {
    withTimeout('60m') {
       def status;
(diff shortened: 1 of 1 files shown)


jobs/deploy-webapp.groovy
```diff
@@ -478,6 +478,20 @@ def _manualSmokeTestCheck(job){
 }
 
 
+def deployFastlyStaging() {
+   withTimeout('10m') {
+      withSecrets.slackAndStackdriverAlertlibOnly() {
+         exec(["make", "-C", "webapp/services/fastly-khanacademy-compute",
```
MiguelCastillo, line 484: Awesome

MiguelCastillo (approved):
This is great!

znd deploys will get a similar treatment.  And call `make build/deploy` in the webapp-build job.  That should complete the jenkins interactions :D

csilvers:
> This is great!
> 
> znd deploys will get a similar treatment. And call `make build/deploy` in the webapp-build job. That should complete the jenkins interactions :dagger: 

znd deploys cannot get this, because you might be deploying a znd the same time someone is doing a deploy, and they can't both use the same fastly 'staging' service.  I think this is an unsolvable problem.  If you want to test a fastly change, the best thing you can do is get in the deploy queue, wait until you're at the front of the line, test on the staging service, and then abort.

In other words, I think _this_ PR finishes our jenkins interactions.  At least for v1.

MiguelCastillo:
> > This is great!
> > znd deploys will get a similar treatment. And call `make build/deploy` in the webapp-build job. That should complete the jenkins interactions 🗡️
> 
> znd deploys cannot get this, because you might be deploying a znd the same time someone is doing a deploy, and they can't both use the same fastly 'staging' service. I think this is an unsolvable problem. If you want to test a fastly change, the best thing you can do is get in the deploy queue, wait until you're at the front of the line, test on the staging service, and then abort.
> 
> In other words, I think _this_ PR finishes our jenkins interactions. At least for v1.

yeah I was just replying to say to hold that thought on znds.  Yeah - this finishes v1 quite nicely.

For testing with znds...  You can deploy to `test` and pass the cookie Im adding to tell fastly vcl which fastly compute service to send requests to.  Im adding such cookie in a PR im sending you on monday :)

PR #241, opened 2024-10-21
Ensure buildmaster code is up to date before deploys

jwbron:
## Summary:
`quickClone` just runs a clone. If the repo is already on disk, the code isn't updated. Instead use the safeSyncToOrigin function. Also change the alert channel name for the sake of reducing noise during the hackathon.

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/jobs/deploy-buildmaster.groovy
+++ b/jobs/deploy-buildmaster.groovy
@@ -35,7 +35,6 @@ def buildAndDeploy() {
 
     dir("buildmaster2") {
       withEnv([
-          "CLOUDSDK_CORE_ACCOUNT=buildmaster-deploy@${params.GCLOUD_PROJECT}.iam.gserviceaccount.com",
           "GCLOUD_PROJECT=${params.GCLOUD_PROJECT}"
       ]) {
         try {


jobs/deploy-buildmaster.groovy
```diff
@@ -27,6 +27,7 @@ currentBuild.displayName = "${currentBuild.displayName} - ${params.GIT_BRANCH} -
 def buildAndDeploy() {
   withTimeout('15m') {
     kaGit.quickClone("git@github.com:Khan/buildmaster2", "buildmaster2", params.GIT_BRANCH);
+    kaGit.safePullInBranch("buildmaster2", params.GIT_BRANCH);
```
csilvers, line 30: I think you just want kaGit.safeSyncToOrigin.  There's no need to do anything fancy here, especially with such a small repo.

jobs/deploy-buildmaster.groovy
```diff
@@ -53,7 +54,7 @@ def buildAndDeploy() {
 }
 
 onMaster('90m') {
-  notify([slack: [channel: '#infrastructure-devops', // we may be deploying Mr. Monkey himself!
+  notify([slack: [channel: '#hack-buildmaster-2024',
                   sender : 'Mr Meta Monkey',
```
csilvers, line 58:
```suggestion
                  sender : 'Mr Meta Monkey',  // we may be deploying Mr. Monkey himself!
```

jobs/deploy-buildmaster.groovy
csilvers: > When we initially tried to use safeSyncToOrigin, the git repo couldn't be found. It seemed to us like safeSyncToOrigin was specifically meant to be used for the webapp repo, is that not the case?

That is not the case.

What is the actual error you got?

> Doing a clone and pull seems like the simplest option.

It's the simplest, but not the most correct.  What happens if somehow the workspace gets dirty?  What if we add submodules?  There are a million things that can go wrong, which we have hardened our scripts to over time.  safeSyncToOrigin is the reliable way to get a particular repo to a particular commit.  As the comments say in kaGit.groovy:
```
// Clone a repo, using our existing workdirs if possible.
// This is very low-level, and does not update submodules!  You probably want
// safeSyncToOrigin, below, instead.
def quickClone(repoToClone, directory, commit) {
   exec(["jenkins-jobs/safe_git.sh", "clone", repoToClone, directory, commit]);
}
```

MiguelCastillo (approved):
This looks good to me.  Im glad to see this getting revived.  If you could please update the description to match your changes before landing, that would be awesome.

PR #242, opened 2024-10-23
Get rid of unnecessary try/catch.

csilvers:
## Summary:
It was setting the status to FAILED, but notify.groovy already does
that so there's no need to do it here too.

I also cleaned up the syntax a bit to be consistent with the rest of
our groovy code, for instance putting semicolons everywhere.

Issue: none

## Test plan:
groovy jobs/deploy-buildmaster.groovy


MiguelCastillo (approved):
great!

PR #243, opened 2024-10-28
Add the concept of a "failure channel" and use it for emergency-rollback.

csilvers:
## Summary:
Sometimes, we send slack messages to a busy or unattended channel,
because most of the time the information isn't that interesting.  But
sometimes it's only not interesting when it works as intended --
e.g. the nightly dev-images update.  But if it fails, we want to know
about it.

This PR introduces the concept of a "failure channel", which if
specified means that when a job fails, we send the slack message
twice: once to the normal channel, but a second time to the failure
channel.  The idea being the failure channel is looked at more
closely.

I use this concept for a few jobs I think would benefit from it:
emergency-rollback -- which has been broken since March and we never
really noticed! -- and the update scripts.

Issue: none

## Test plan:
Fingers crossed.  I did run
```
groovy vars/notify.groovy
groovy jobs/emergency-rollback.groovy
[and likewise for the other changed files]
```

### DIFF
--- a/jobs/emergency-rollback.groovy
+++ b/jobs/emergency-rollback.groovy
@@ -144,6 +144,7 @@ def doRollback() {
 
 onMaster('1h') {
    notify([slack: [channel: '#1s-and-0s-deploys',
+                   failureChannel: '#infrastructure-platform',
                    sender: 'Mr Monkey',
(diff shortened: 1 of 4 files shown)


nathanjd (approved):
Seems reasonable to me.

PR #246, opened 2024-11-01
Update buildmaster endpoint to the new cloud run deployment

jwbron:
## Summary:
This points to the new cloud run deployment of buildmaster

Issue: INFRA-10447

## Test plan:

### DIFF
--- a/vars/buildmaster.groovy
+++ b/vars/buildmaster.groovy
@@ -70,7 +70,7 @@ def _makeHttpRequestAndAlert(resource, httpMode, params) {
                                 maskValue: true]],
                httpMode: httpMode,
                requestBody: new JsonBuilder(params).toString(),
-               url: "https://buildmaster.khanacademy.org/${resource}");
+               url: "https://buildmaster-526011289882.us-central1.run.app/${resource}");
             notify.log("Buildmaster request: ${resource}", [
                resource: resource,
                response: response,


nathanjd (approved):

vars/buildmaster.groovy
```diff
@@ -70,7 +70,7 @@ def _makeHttpRequestAndAlert(resource, httpMode, params) {
                                 maskValue: true]],
                httpMode: httpMode,
                requestBody: new JsonBuilder(params).toString(),
-               url: "https://buildmaster.khanacademy.org/${resource}");
+               url: "https://buildmaster-526011289882.us-central1.run.app/${resource}");
```
csilvers, line 73: Where did this name come from?  My guess is we want another degree of indirection here.

For instance, should we change buildmaster.ka.org to point to this url?  Or set up an alias in cloud-run?  Where did this number 526011289882 come from?

csilvers: OK!

MiguelCastillo: I know this landed.  It would be good to include details i code about where these different things come from.  Like the cloudrun buildmaster hostname and such.  Like adding a comment that tells future us 

> And you can find the endpoint defined here: https://console.cloud.google.com/run/detail/us-central1/buildmaster/metrics?project=khan-internal-services
> 526011289882 is the project id for khan-internal-services so it will never change.

MiguelCastillo: > Miguel suggested that we get rid of the buildmaster.ka.org endpoint. This endpoint is stable for the given cloud run service name, project, and region.

We can configure that hostname.  My main motivation is to allow us to transition to the new system without having to wait for dns propagation of sorts.  And Im just a bit hesitant to have configurations in places not every has access too or somehow obscure how it's setup.

But we can reassess this choice once we are all migrated!

PR #248, opened 2024-11-08
Clean up the buildmaster tags every week.

csilvers:
## Summary:
We have 55,000 of these!  They were meant to be short-lived (like a
few minutes long) tags, but we never wrote the code to prune them.
Let's do that now.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1731094743406639

## Test plan:
Fingers crossed.  I ran the grep part manually and saw it exclude tags
from the last 7 days.

### DIFF
--- a/weekly-maintenance-jobs.sh
+++ b/weekly-maintenance-jobs.sh
@@ -79,6 +79,26 @@ clean_invalid_branches() {
     done
 }
 
+# The merge-branches jenkins job creates a git-tag to make sure the
+# merged code isn't gc-ed by Jenkins.  It only needs to survive for
+# the length of a deploy: maybe an hour or so.  We keep these tags
+# around a week.
+prune_buildmaster_tags() {
+    (
+        cd webapp
+        # We delete tags 100 at a time to avoid overwhelming github.
+        git tag -l 'buildmaster-*' \
+            | grep -v \
+                   -e buildmaster-'[0-9]*'-$(date -d "today" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -1 day" +%Y%m%d) \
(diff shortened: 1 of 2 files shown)


MiguelCastillo (approved):
Great! thank you, craig.

weekly-maintenance-jobs.sh
```diff
@@ -79,6 +79,26 @@ clean_invalid_branches() {
...
+        git tag -l 'buildmaster-*' \
+            | grep -v \
+                   -e buildmaster-'[0-9]*'-$(date -d "today" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -1 day" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -2 days" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -3 days" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -4 days" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -5 days" +%Y%m%d) \
+                   -e buildmaster-'[0-9]*'-$(date -d "today -6 days" +%Y%m%d) \
+            | xargs -n 100 git push --no-verify --delete origin
```
MiguelCastillo, line 99: at a glance this seems reasonable.  but it is tricky to know if it's the perfect incantation.  `date -d "today -6 days" +%Y%m%d` doesn't work mac.  I do believe it does work ubuntu tho.

csilvers: Yeah, I do have lots of linux-specific stuff in these jobs.

PR #250, opened 2024-11-13
Add FASTLY_SERVICE param to E2E test job

kevinb-khan:
## Summary:
We want to be able to easily run E2E jobs using different Fastly services.  This PR adds a new param to the E2E test job to support this.  It requires https://github.com/Khan/webapp/pull/27243 in order to work correctly.

Issue: None

## Test plan:
- land this PR
- run the E2E tests on 'route-e2e-tests-to-compute' using 'PROD [Compute]' from the dropdown

PR #251, opened 2024-11-13
[🔥AUDIT🔥] Pass FASTLY_COMPUTE_ENVIRONMENT_COOKIE to lambdatest

kevinb-khan:
🖍 _This is an audit!_ 🖍

## Summary:
@miguelcastillo caught a mistake I made in #250.  I was passing the service ID instead of the cookie value.  This PR fixes that and adds a link to the Confluence page with the cookie values which is our source of truth for these values.

The corresponding webapp changes can be found in https://github.com/Khan/webapp/pull/27243.

Issue: None

## Test plan:
- land
- start a job and stop it (this will force jenkins to update)
- start a new job that runs tests on the `route-e2e-tests-to-compute` branch in webapp and sets the `FASTLY_SERVICE` param to `PROD [Compute]`

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -181,19 +181,21 @@ def runLambdaTest() {
    // in the LambdaTest build.
    def e2eEnv = E2E_URL == "https://www.khanacademy.org" ? "prod" : "preprod";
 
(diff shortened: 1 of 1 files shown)


csilvers (commented):
:+1:

PR #254, opened 2024-12-06
Add temporary Cypress Cloud integration

jandrade:
## Summary:

Adds a temporary jenkins job to run e2e tests on Cypress Cloud. This job will be
removed once the tests are running successfully on Cypress Cloud.

The goal is to use parallel workers in Jenkins directly to run concurrent
cypress processes. This will allow us evaluate the performance of the tests
and the feasibility of switching to Cypress Cloud.

NOTE: I used the previous `e2e-test.groovy` template as reference: https://github.com/Khan/jenkins-jobs/blob/fc496e120f19c439348ae4136a0f92c909b5ab2c/jobs/e2e-test.groovy

Issue: XXX-XXXX

## Test plan:

Make sure that the `CYPRESS_GIT_REVISION` parameter is set to the correct branch:
`cycloud-skipped`.

1. Run the job `e2e-test-cycloud` in Jenkins.
2. Verify that the tests are running on Cypress Cloud.

### DIFF
--- a/jobs/e2e-test-cycloud.groovy
+++ b/jobs/e2e-test-cycloud.groovy
@@ -202,7 +202,7 @@ def runE2ETests(workerId) {
 
    // NOTE: We hard-code the values here as it is an experiment that will be
    // removed soon.
-   def runE2ETestsArgs = ["CYPRESS_PROJECT_ID=2c1iwj && CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5 && ./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
+   def runE2ETestsArgs = ["env", "CYPRESS_PROJECT_ID=2c1iwj", "CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5", "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
 
    dir('webapp/services/static') {
       exec(runE2ETestsArgs);
--- a/jobs/e2e-test-cycloud.groovy
+++ b/jobs/e2e-test-cycloud.groovy
@@ -112,12 +112,6 @@ for more information.""",
 // We set these to real values first thing below; but we do it within
 // the notify() so if there's an error setting them we notify on slack.
 NUM_WORKER_MACHINES = null;
-// Used to tell whether all the test-workers raised an exception.
-WORKERS_RAISING_EXCEPTIONS = 0;
-public class TestFailed extends Exception {}  // for use with the above
-// Used to make sure we finish our job as soon as all tests are run.
-public class TestsAreDone extends Exception {}
-
 
 // Override the build name by the info that is passed in (from buildmaster).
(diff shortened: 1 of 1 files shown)


csilvers:
Can you say what file you used as a template for this?  I suspect it will be easier for me to look at a diff between that file and this, than to review this file from scratch.

jobs/e2e-test-cycloud.groovy
```diff
@@ -0,0 +1,230 @@
...
+def runE2ETests(workerId) {
+   echo("Starting e2e tests for worker ${workerId}");
+
+   // Determine which environment we're running against, so we can provide a tag
+   // in the LambdaTest build.
+   def e2eEnv = E2E_URL == "https://www.khanacademy.org" ? "prod" : "preprod";
+
+   // NOTE: We hard-code the values here as it is an experiment that will be
+   // removed soon.
+   def runE2ETestsArgs = ["CYPRESS_PROJECT_ID=2c1iwj && CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5 && ./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
```
csilvers, line 205:
```suggestion
   def runE2ETestsArgs = ["env", "CYPRESS_PROJECT_ID=2c1iwj", "CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5", "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
```

csilvers: What you have now won't work, it will try to find a binary with the name of that entire string.

An alternative to using `env` would be to set the environment in jenkins via `withEnv`.

jobs/e2e-test-cycloud.groovy
```diff
@@ -0,0 +1,230 @@
...
+      };
+   }
+   parallel(jobs);
+}
+
+
+def runTests() {
+   try {
+     runAllTestClients();
+    } catch (TestsAreDone e) {
```
csilvers, line 189: You don't make use of this -- you never throw it -- and can just remove it.  Likewise TestFailed.

csilvers (approved):
I'm of two minds about this PR.  You clearly took an existing groovy file and modified it to work for your needs, but in the process you left in a bunch of stuff that doesn't really apply anymore, which makes the job pretty confusing.  For testing purposes, this file doesn't need to be maintainable, so I'm approving, but when the time comes to replace the existing e2e-test.groovy, we'll need to do some work to get it a bit more maintainable!

jandrade:
> For testing purposes, this file doesn't need to be maintainable, so I'm approving, but when the time comes to replace the existing e2e-test.groovy, we'll need to do some work to get it a bit more maintainable!

@csilvers 100% agree with this. The goal for now is the see if it works, as we are not fully decided if we are going to migrate to Cypress Cloud yet. There are still some things that we want to test before making a decision + we also need to get the budget approved for this change to happen.

If we decide to adopt this platform, for sure we are going to be updating this job to cover all the cases (including failures, reporting, etc).

jobs/e2e-test-cycloud.groovy
jandrade: Thanks! I've removed that logic (including the failures part).

PR #259, opened 2025-01-22
Don't call yarn directly in jenkins-jobs.

jeresig:
## Summary:
This is a follow-up to:
https://github.com/Khan/webapp/pull/28829

I changed the few places where we install packages to use `make npm_deps` instead. The two places where we were calling a yarn script I now use the new shell script for this, instead. This is all in service of moving off of yarn v1 to pnpm v10 instead.

Issue: FEI-6143

## Test plan:
Will need to deploy the webapp changes first. But I'm a bit unclear on how to best test this!

### DIFF
--- a/jobs/demo-district-update-pass.groovy
+++ b/jobs/demo-district-update-pass.groovy
@@ -40,13 +40,13 @@ BUILD_NAME = "build demo-district-password-update";
 def _setupWebapp() {
    kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", params.CYPRESS_GIT_REVISION);
    dir("webapp/services/static") {
-      sh("yarn install");
+      sh("make npm_deps");
    }
 }
 
 def runScript() {
    def workDir = "webapp/services/static/javascript/districts-package/__e2e-tests__";
-   def runCypressArgs = ["yarn",
+   def runCypressArgs = ["./dev/tools/run_pkg_script.sh",
                          "cypress",
                          "run", 
                          "--spec",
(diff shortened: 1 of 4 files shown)


kevinb-khan (approved):
LGTM 🎉

csilvers (approved):
It would be great if you updated the commit message to make it clearer that the reason for doing this is to enable an eventual transition away from yarn.

Assuming that's the motivation, this looks good to me!  I knew `make` would win the day in the end. :-)

> Will need to deploy the webapp changes first. But I'm a bit unclear on how to best test this!

The easiest way is once your webapp changes are in master.  Then you can pick the latest run for each of these jobs and hit `replay`, e.g. https://jenkins.khanacademy.org/job/deploy/job/e2e-test/70455/replay/.  Then replace the entire "main script" section with your new code (as of this PR), and run it.

It looks like demo-district and e2e-test-cycloud are not currently passing, so I don't know that you can test them.  For webapp-maintenance, I think it's fine to just wait until the weekend and see how it does; it's not the end of the world if it fails.  But you can also do the "replay" thing for that if you'd like.

csilvers:
If you want to test before your webapp changes are deployed, you can do that too.  It'll take two steps.  First, you run the job (say, e2e-test) on the branch you want to test; that is, you set `GIT_REVISION` to point to your branch.  Once that's started -- you can let it run to completion, or abort it -- you can select that job and hit "replay" and put in your new code.

jobs/demo-district-update-pass.groovy
```diff
@@ -40,13 +40,13 @@ BUILD_NAME = "build demo-district-password-update";
 def _setupWebapp() {
    kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", params.CYPRESS_GIT_REVISION);
    dir("webapp/services/static") {
-      sh("yarn install");
+      sh("make npm_deps");
```
rgagarinr, line 43: Thanks, John, for fixing it!

PR #260, opened 2025-01-22
Integrate e2e-test-cycloud job with Buildmaster

rgagarinr:
Issue: FEI-6125

Summary:
This draft PR introduces a Cypress Cloud job that runs in parallel with the existing Lambda e2e-test job. It also includes a Mocha implementation for each worker. (Note: The Mocha implementation requires updates with `yarn`, which can be addressed in a subsequent PR.)

The primary goal is to utilize Jenkins stash functionality to capture and merge all junit reports into a single file, enabling streamlined processing.


jobs/e2e-test-cycloud.groovy
```diff
@@ -166,15 +166,25 @@ def runAllTestClients() {
...
+      "--reporter mocha-junit-reporter",
+      "--reporter-options mochaFile=${junitFile}"
+   ];
 
    dir('webapp/services/static') {
+      // We build the secrets first, so that we can create test users in our
+      // tests.
+      sh("yarn cypress:secrets");
       exec(runE2ETestsArgs);
    }
```
rgagarinr, line 189: The idea is stashing of individual worker results `stash includes: "${resultsDir}/junit-${workerId}.xml", name: "worker-${workerId}-results";`

jobs/e2e-test.groovy
```diff
@@ -338,7 +359,11 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
...
          // for actual smoketest-code errors in analyzeResults(), below.
-         runLambdaTest();
+         parallel([
+               "run-lambdatest": { runLambdaTest(); },
+               "test-cycloud": { runCypressCloud(); },
+            ]);
+         // runLambdaTest();
       }
 
       stage("Analyzing results") {
```
rgagarinr, line 369:
```
// Collect results from the triggered job
        stage("Collect Cypress Results") {
            def resultsDir = "results";
            def mergedResults = "${resultsDir}/junit-merged.xml";

            sh("mkdir -p ${resultsDir}");

            // Unstash results from all workers
            for (int i = 0; i < params.NUM_WORKER_MACHINES.toInteger(); i++) {
                unstash name: "worker-${i}-results";
            }

            // Merge results into a single JUnit XML file
            sh("""
                find ${resultsDir} -name '*.xml' -print0 | xargs -0 cat > ${mergedResults}
            """);

            echo("Merged results written to ${mergedResults}");
        }
```

jobs/e2e-test-cycloud.groovy
```diff
@@ -166,15 +166,25 @@ def runAllTestClients() {
...
+      "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts",
+      "--cyConfig baseUrl=\"${E2E_URL}\"",
+      "--reporter mocha-junit-reporter",
+      "--reporter-options mochaFile=${junitFile}"
+   ];
 
    dir('webapp/services/static') {
+      // We build the secrets first, so that we can create test users in our
+      // tests.
+      sh("yarn cypress:secrets");
```
csilvers, line 187: Cc @jeresig -- is this something we need to be doing some other way, now?

jobs/e2e-test-cycloud.groovy
```diff
@@ -166,15 +166,25 @@ def runAllTestClients() {
 def runE2ETests(workerId) {
```
csilvers, line 166: What directory are we in when this function is run?

jobs/e2e-test-cycloud.groovy
jeresig: Yes - this will need to use the new wrapper script I added here: https://github.com/Khan/webapp/pull/28829 good catch! I'm going to be landing the script to master today.

jobs/e2e-test-cycloud.groovy
```diff
@@ -166,15 +166,25 @@ def runAllTestClients() {
...
+   // Define which environment we're running against, and setting up junit report
    def e2eEnv = E2E_URL == "https://www.khanacademy.org" ? "prod" : "preprod";
-
-   // NOTE: We hard-code the values here as it is an experiment that will be
-   // removed soon.
-   def runE2ETestsArgs = ["env", "CYPRESS_PROJECT_ID=2c1iwj", "CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5", "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
+   def resultsDir = "results"; // Directory to store results
+   def junitFile = "${resultsDir}/junit-${workerId}.xml";
+
+   // Ensure the results directory exists
```
csilvers, line 174: Don't you also need to clear out any old xml files that might be there?

jobs/e2e-test-cycloud.groovy
csilvers: Are you planning on doing that in this PR, or a follow-up PR?

csilvers (commented):
This seems pretty safe to start off with, I appreciate you being careful about the new job not interfering with the existing job.  I want to make sure I understand about the `rm` first before approving though.

jobs/e2e-test-cycloud.groovy
```diff
@@ -166,15 +166,25 @@ def runAllTestClients() {
...
+   sh("mkdir -p ${resultsDir}");
+   
+   def runE2ETestsArgs = [
+      "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts",
+      "--cyConfig baseUrl=\"${E2E_URL}\"",
+      "--reporter mocha-junit-reporter",
+      "--reporter-options mochaFile=${junitFile}"
```
somewhatabstract, line 181: **note:** This invocation will need to change. These args aren't supported by the runner and I'm not sure we need to provide them like this. I think most of this can be calculated internally to the code.

jobs/e2e-test.groovy
somewhatabstract: Is this how results get merged for unit tests too? If not, what's different and why?

jobs/e2e-test-cycloud.groovy
somewhatabstract: In addition, we're not going to do junit right away. That's a stretch goal once we have things working.

jobs/e2e-test-cycloud.groovy
somewhatabstract: We don't need to call `yarn cypress:secrets` at all. Secrets are handled as passed environment variables, rather than a JSON file.

jobs/e2e-test-cycloud.groovy
rgagarinr: Sorry for the confusion, but I think yes, we will work on the junit implementation after this PR.

jobs/e2e-test.groovy
rgagarinr: I think it’s a slightly different alternative way of how we might collect junit results - maybe more straightforward.

jobs/e2e-test-cycloud.groovy
rgagarinr: If I’m not mistaken, the script executes the runE2ETests function within the webapp/services/static directory. This is defined in the dir('webapp/services/static') block in the parent function (runAllTestClients).

jobs/e2e-test-cycloud.groovy
rgagarinr: Oh, thanks! Yeah, we need to clean up the old xml files first to avoid conflicts.

jobs/e2e-test-cycloud.groovy
csilvers: OK, you don't want to create the tmpdir in webapp/services/static/results -- you want it to be at the top level.  According to my web research, you can do:
```
resultsDir = "${WORKSPACE}/results"; // Directory to store results
```
but you'll have to test that.

Also, "results" is a very generic name.  How about something like "junit-output/"?

rgagarinr: Thanks, Craig! {WORKSPACE} works, yeah and I’ll rename `results`! cc @michaelmendoza42

PR #261, opened 2025-01-23
Add job to validate exercise IDs in assessments

timmcca-be:
## Summary:
Our interim assessments contain exercise IDs that we use to look up content, and we need to be alerted if any of those IDs become invalid. This job checks for invalid exercise IDs and alerts us if it finds any.

Issue: INFRA-10509

## Test plan:
Verify that the job ran correctly [in Jenkins](https://jenkins.khanacademy.org/job/misc/job/check-assessment-exercise-ids/). Note that the failure is intended - I made a test configuration invalid to ensure the alert would work.

### DIFF
--- a/jobs/check-assessment-exercise-ids.groovy
+++ b/jobs/check-assessment-exercise-ids.groovy
@@ -7,7 +7,9 @@
 import org.khanacademy.Setup;
 
 
-new Setup(steps).apply();
+new Setup(steps).addCronSchedule(
+   '0 * * * *' // Run every hour
+).apply();
 
 
 def _setupWebapp() {
@@ -20,7 +22,7 @@ def _setupWebapp() {
 def runScript() {
    withTimeout('1h') {
       dir("webapp") {
-         sh("go run services/assessments/cmd/check_assessments_exercise_ids/main.go");
+         exec(["go", "run", "services/assessments/cmd/check_assessments_exercise_ids/main.go"]);
       }
    }
 }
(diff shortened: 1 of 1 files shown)


jobs/check-assessment-exercise-ids.groovy
```diff
@@ -0,0 +1,41 @@
+// Loads all the assessment configs from datastore and checks to make sure that
+// their ExerciseIDs are valid (i.e., that there are published exercises with
+// those IDs).
+
+@Library("kautils")
+// Classes we use, under jenkins-jobs/src/.
+import org.khanacademy.Setup;
+
+
+new Setup(steps).apply();
```
csilvers, line 10: You want to run this as a cronjob, right?  So you need to add a cron step here.  Though maybe you're planning to do that later once you confirm this works?  I'd probably do it from the beginning, tbh.  It's pretty straightforward what you're doing and all the work is in the go binary, so this won't take a lot of iterating.

jobs/check-assessment-exercise-ids.groovy
```diff
@@ -0,0 +1,41 @@
...
+   withTimeout('1h') {
+      kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", "master");
+   }
+}
+
+
+def runScript() {
+   withTimeout('1h') {
+      dir("webapp") {
+         sh("go run services/assessments/cmd/check_assessments_exercise_ids/main.go");
```
csilvers, line 23: This is fine, and certainly very common in our jenkins code, but I think it's a bit safer to do exec:
```
    exec(["go", "run", "services/..."]);
```
It's totally up to you.

csilvers (approved):
Looks great!

jobs/check-assessment-exercise-ids.groovy
timmcca-be: Ah okay, I thought I was supposed to configure that in Jenkins! Will add it now

PR #263, opened 2025-02-06
update e2e-test-cycloud to merge and analyze test results

michaelmendoza42:
## Summary:

update e2e-test-cycloud to merge and analyze test results
in conjunction with webapp update: https://github.com/Khan/webapp/pull/29302

Issue: FEI-6124

## Test plan:
ran script on test job and the jobs were reporting successfully

### DIFF
--- a/jobs/e2e-test-cycloud.groovy
+++ b/jobs/e2e-test-cycloud.groovy
@@ -125,6 +125,8 @@ BUILD_NAME = "build e2e-cypress-test #${env.BUILD_NUMBER} (${E2E_URL}: ${params.
 
 // GIT_SHA1 is the sha1 for CYPRESS_GIT_REVISION.
 GIT_SHA1 = null;
+REPORT_DIR = "webapp/services/static/cypress/results/mochawesome"
+REPORT_NAME = "mochawesome-combined.json"
 
 // We have a dedicated set of workers for the second smoke test.
 WORKER_TYPE = (params.USE_FIRSTINQUEUE_WORKERS
@@ -132,13 +134,11 @@ WORKER_TYPE = (params.USE_FIRSTINQUEUE_WORKERS
 
 def initializeGlobals() {
    NUM_WORKER_MACHINES = params.NUM_WORKER_MACHINES.toInteger();
+   GIT_SHA1 = kaGit.resolveCommitish("git@github.com:Khan/webapp",
+           params.CYPRESS_GIT_REVISION);
 }
 
-
 def _setupWebapp() {
-   GIT_SHA1 = kaGit.resolveCommitish("git@github.com:Khan/webapp",
-                                        params.CYPRESS_GIT_REVISION);
-
    kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_SHA1);
 
    dir("webapp/services/static") {
@@ -152,14 +152,17 @@ def runAllTestClients() {
    for (i = 0; i < NUM_WORKER_MACHINES; i++) {
       def workerId = i;  // avoid scoping problems
       jobs["e2e-test-${workerId}"] = {
-         swallowExceptions({
+         stage("e2e-worker-${workerId}") {
             onWorker(WORKER_TYPE, '2h') {
-                _setupWebapp();
-                runE2ETests(workerId);
+               _setupWebapp()
+               runE2ETests(workerId)
+               dir("${REPORT_DIR}") {
+                  stash includes: "*.json", name: "worker-${workerId}-reports"
+               }
             }
-         }, onException);
-      };
-   }
+         }
+      }
+   };
    parallel(jobs);
 }
 
@@ -170,12 +173,20 @@ def runE2ETests(workerId) {
    // in the LambdaTest build.
    def e2eEnv = E2E_URL == "https://www.khanacademy.org" ? "prod" : "preprod";
 
-   // NOTE: We hard-code the values here as it is an experiment that will be
-   // removed soon.
-   def runE2ETestsArgs = ["env", "CYPRESS_PROJECT_ID=2c1iwj", "CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5", "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
+   def runE2ETestsArgs = [
+      "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts",
+      "--cyConfig reporter=mochawesome"
+   ];
 
    dir('webapp/services/static') {
-      exec(runE2ETestsArgs);
+      // We build the secrets first, so that we can create test users in our
+      // tests.
+      sh("yarn cypress:secrets");
+
+      withEnv(["CYPRESS_PROJECT_ID=2c1iwj",
+               "CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5"]) {
+         exec(runE2ETestsArgs);
+      }
    }
 }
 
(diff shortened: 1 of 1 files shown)


jobs/e2e-test-cycloud.groovy
```diff
@@ -17,94 +18,94 @@ new Setup(steps
...
-   "https://www.khanacademy.org"
+        "URL",
+        "The url-base to run these tests against.",
+        "https://www.khanacademy.org"
 
 )
-.addChoiceParam(
-   "TEST_TYPE",
-   """IGNORE: This is a dummy parameter that is only here to avoid breaking the
+        .addChoiceParam(
```
csilvers, line 29: This should be on one line: `).addChoiceParam(`.

jobs/e2e-test-cycloud.groovy
```diff
@@ -17,94 +18,94 @@ new Setup(steps
...
+                "TEST_TYPE",
+                """IGNORE: This is a dummy parameter that is only here to avoid breaking the
    communication with buildmaster""",
-   ["all", "deploy", "custom"]
+                ["all", "deploy", "custom"]
 
-).addStringParam(
-   "TESTS_TO_RUN",
-   """IGNORE: This is a dummy parameter that is only here to avoid breaking the
+        ).addStringParam(
```
csilvers, line 35: Not sure why this is indented, but it shouldn't be.  In general, I wonder if you want to revert all the whitespace changes here.

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
    for (i = 0; i < NUM_WORKER_MACHINES; i++) {
       def workerId = i;  // avoid scoping problems
       jobs["e2e-test-${workerId}"] = {
-         swallowExceptions({
```
csilvers, line 155: Do you not want to swallow exceptions anymore?

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
...
-   def runE2ETestsArgs = ["env", "CYPRESS_PROJECT_ID=2c1iwj", "CYPRESS_RECORD_KEY=4c530cf3-79e5-44b5-aedb-f6f017f38cb5", "./dev/cypress/e2e/tools/start-cypress-cloud-run.ts"];
+   def runE2ETestsArgs = [
+           "./dev/cypress/e2e/tools/start-cy-cloud-run.ts",
+           "--url=${E2E_URL}",
+           "--name=${BUILD_NAME}",
+   ];
 
    dir('webapp/services/static') {
-      exec(runE2ETestsArgs);
+      exec(runE2ETestsArgs);}
```
csilvers, line 184:
```suggestion
      exec(runE2ETestsArgs);
```

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
...
+   }
+}
+
+def unstashReports() {
+   def jsonFolders = []
+   dir("${REPORT_DIR}") {
+      for (i = 0; i < NUM_WORKER_MACHINES; i++) {
+         sh "mkdir -p ./${i}"
```
csilvers, line 192:
```suggestion
         exec(["mkdir", "-p", "${i}"]);
```
But I think you need to do an `rmdir` first.  Otherwise, the results from a previous run will still be sticking around here.

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
...
+         jsonFolders.add("${i}")
+         dir("./${i}") {
+            sh "pwd"
```
csilvers, line 195:
```suggestion
            sh("pwd");
```
We always use parens and semicolons in our groovy, even though the language doesn't require it.  (Here and below.)

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
...
+            try {
+               unstash "worker-${i}-reports"
+            } catch (Exception e) {
+               echo "Failed to unstash worker-${i}-reports: ${e.getMessage()}"
```
csilvers, line 199: You want to continue in this case?

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
...
+   ];
 
    dir('webapp/services/static') {
-      exec(runE2ETestsArgs);
+      exec(runE2ETestsArgs);}
+   }
+}
+
+def unstashReports() {
+   def jsonFolders = []
```
csilvers, line 189:
```suggestion
   def jsonFolders = [];
```
(You should check you have semicolons everywhere.)

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
...
+   }
+
+   // report-merged-results.ts is a new file
+   kaGit.safePullInBranch("webapp/services/static/dev/cypress/e2e/tools", params.CYPRESS_GIT_REVISION);
+
+   dir ('webapp/services/static') {
+      sh "ls ./dev/cypress/e2e/tools"
+      catchError(buildResult: "UNSTABLE", stageResult: "UNSTABLE",
+              message: "There were test failures!") {
+         sh "npx --yes tsx ./dev/cypress/e2e/tools/report-merged-results.ts ${foldersList.join(' ')}"
```
csilvers, line 223: Definitely use `exec` here, and I think you can just do `[...] + foldersList`.

csilvers (changes requested):
I like the general idea of stashing and unstashing!  Request Changes because there are some style issues that are keeping me from really understanding the code.  In particular, I don't understand exactly what happens when various parts of the system have framework errors -- when you can't unstash, or when a test you expect to be in the json files isn't there.

jobs/e2e-test-cycloud.groovy
```diff
@@ -17,94 +18,94 @@ new Setup(steps
 // We do a lot of e2e-test runs, and QA would like to be able to see details
 // for a bit longer.
 ).resetNumBuildsToKeep(
-   350,
+        350,
```
somewhatabstract, line 21: **nit:** It looks like your indent settings were wrong and this got formatted unnecessarily across the file. I would fix this to simplify the review.

jobs/e2e-test-cycloud.groovy
```diff
@@ -152,30 +154,74 @@ def runAllTestClients() {
    for (i = 0; i < NUM_WORKER_MACHINES; i++) {
       def workerId = i;  // avoid scoping problems
       jobs["e2e-test-${workerId}"] = {
-         swallowExceptions({
+         stage("e2e-worker-${workerId}") {
             onWorker(WORKER_TYPE, '2h') {
-                _setupWebapp();
-                runE2ETests(workerId);
+               _setupWebapp()
+               runE2ETests(workerId)
```
somewhatabstract, line 160: **suggestion:** You'll want to make sure old results are tidied up from genfiles before you run, I think.

jobs/e2e-test-cycloud.groovy
somewhatabstract: Yes, I believe something in this job somewhere needs to clean up old results files so that we're only reporting on this specific run.

jobs/e2e-test-cycloud.groovy
somewhatabstract: **suggestion:** Per Craig's question, I think we have to fail if we can't get our results.

somewhatabstract:
Does this include the changes @rgagarinr has ready to ensure we will run side-by-side with the LT version at first? If not, should that be a part of this PR so that it all lands at the same time or should that be separate?

jobs/e2e-test-cycloud.groovy
michaelmendoza42: yeah i'll revert them. didn't realize it until i saw it here.

jobs/e2e-test-cycloud.groovy
michaelmendoza42: ah yeah i think this is a remnant from debugging

rgagarinr:
> Does this include the changes @rgagarinr has ready to ensure we will run side-by-side with the LT version at first? If not, should that be a part of this PR so that it all lands at the same time or should that be separate?

This [PR](https://github.com/Khan/jenkins-jobs/pull/264) has already been approved we can merge both.

jobs/e2e-test-cycloud.groovy
csilvers: I think nothing runs because you haven't defined `onException` in this script.  You should copy it from webapp-test if you want it.  In general, you should think about what you want to actually happen if one of the workers throws an exception for some reason.  In webapp-test, we depend on analysis of the junit xml files to figure out what worked and what didn't, so swallowing exceptions is correct.  But it's not clear to me if that's what you're doing here or not.

csilvers (approved):
This seems right to me!  But I would do the following test-plan before landing:

Go to https://jenkins.khanacademy.org/job/misc/job/e2e-test-cycloud/lastSuccessfulBuild/, click on the "replay" button on the side, and replace all the text in the `main` textbox with your new code here.  And then run it and see how it goes.  It may not work if there are files you need in webapp that don't exist at the commit being tested, but if that turns out not to be an issue you can have confidence this is doing what you want!

Maybe you already did this, and just need to update the test plan.

P.S. why is this in the `misc` folder instead of `deploy`?  Do you intend it to be part of the deploy process one day?

PR #266, opened 2025-02-11
[🔥AUDIT🔥] change path of cycloud job now that it is added to deploy folder

michaelmendoza42:
🖍 _This is an audit!_ 🖍

## Summary:
changed cycloud call in e2e-test to use the new one created in deploys `e2e-test-cycloud`

Issue: none
## Test plan:
none


csilvers (approved):
Yay!

PR #267, opened 2025-02-13
Add the ability to deploy to the fastly kastatic service.

csilvers:
## Summary:
I'm moving that service to being configured via yaml instead of
configured via the UI.  So we'll want to be able to deploy it, just
like fastly-vcl.  This PR does that.

Issue: https://khanacademy.atlassian.net/browse/INFRA-10537

## Test plan:
Will try this once it's landed, sadly I can't do it before since it
updates the parameters.

### DIFF
--- a/jobs/deploy-fastly.groovy
+++ b/jobs/deploy-fastly.groovy
@@ -48,22 +48,23 @@ phabricator/diff/&lt;id&gt; (using the latest ID from the diff's "history" tab o
     "TARGET",
     """\
 <ul>
-  <li> <b>test</b>: https://www-boxes.khanacademy.org
+  <li> <b>test</b>: https://www-test.khanacademy.org, www-boxes.khanacademy.org
   <li> <b>prod</b>: https://www.khanacademy.org
-  <li> <b>staging</b>: TODO [compute@edge only]
+  <li> <b>staging</b>: https://www-staging.khanacademy.org [compute@edge only]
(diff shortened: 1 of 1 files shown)


jobs/deploy-fastly.groovy
```diff
@@ -48,22 +48,23 @@ phabricator/diff/&lt;id&gt; (using the latest ID from the diff's "history" tab o
...
 ).addChoiceParam(
-    "LANGUAGE",
+    "SERVICE",
     """\
 <ul>
-  <li> <b>vcl</b>
-  <li> <b>compute@edge</b>
+  <li> <b>fastly-vcl</b>
+  <li> <b>fastly-compute</b>
+  <li> <b>kastatic</b>
```
kevinb-khan, line 64: Does it matter what target people select when choosing this service?

kevinb-khan (approved):
The changes seem reasonable to me, but jenkins-jobs is an area of weakness for me so you may want to get another review.  Were there webapp changes to `services/fastly-kastatic` necessary to make this work?  If so, could you link to them in the summary of this PR?

jobs/deploy-fastly.groovy
csilvers: No.  In general, it would be great if we could modify the list of targets based on the value of SERVICE, but that's not functionality that jenkins provides.

PR #268, opened 2025-02-18
Remove lambdatest from e2e-test.groovy

michaelmendoza42:
## Summary:
<Add your summary here>

Issue: FEI-6126

## Test plan:
none


jobs/e2e-test.groovy
```diff
@@ -233,110 +182,11 @@ def runCypressCloud(){
           ],
           propagate: false,
           // The pipeline will NOT wait for this job to complete to avoid
-          // blocking the main pipeline (runLambdaTest).
+          // blocking the main pipeline (runCypressCloud).
```
somewhatabstract, line 185: **praise:** Good catch.

jobs/e2e-test.groovy
```diff
@@ -354,20 +204,10 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
...
-            ]);
-         // runLambdaTest();
-      }
-
-      stage("Analyzing results") {
-         analyzeResults();
+         // Note: runCypressCloud() analyzes the results so it will post the correct RC
+         // The slack notificaiton is done via the Cypress Cloud integration and
+         // is no longer needed here, also
+         runCypressCloud();
```
somewhatabstract, line 210: **question:** Which part of this communicates test failures to buildmaster?

Right now, our builds fail if lambdatest has failing tests. I presume we need to do something to ensure the same happens when CyCloud tests fail, but I don't see that change here. Is that in a different place?

jobs/e2e-test.groovy
```diff
@@ -233,110 +182,11 @@ def runCypressCloud(){
           ],
           propagate: false,
           // The pipeline will NOT wait for this job to complete to avoid
```
somewhatabstract, line 184: **question:** Don't we want the pipeline to wait for this job now?

jobs/e2e-test.groovy
```diff
@@ -233,110 +182,11 @@ def runCypressCloud(){
...
-          // blocking the main pipeline (runLambdaTest).
+          // blocking the main pipeline (runCypressCloud).
           wait: false,
```
somewhatabstract, line 186: **question:** Don't we want the pipeline to wait for this job now?

somewhatabstract (commented):
This looks good but I think it's incomplete. We need to make sure that deploys will wait on these tests and that the E2E steps in buildmaster fail if tests fail. I don't see changes here that make that happen.

somewhatabstract (changes requested):
This looks good but I think it's incomplete. We need to make sure that deploys will wait on these tests and that the E2E steps in buildmaster fail if tests fail. I don't see changes here that make that happen.

jobs/e2e-test.groovy
nathanjd, line 210: Seems nice that we don't have to manage the slack notifications anymore.

Was something supposed to come after

> , also

?

jobs/e2e-test.groovy
nathanjd, line 185: So this job now only exists to spawn another job? It seems like we should either move all the code from `e2e-test-cycloud.groovy` into this file (easy) or update all references to the new job (not as easy as it involves buildmaster changes).

Are there other dependencies that expect the the job that runs cypress to be named `e2e-test-cycloud`?

jobs/e2e-test.groovy
nathanjd: Yes I think we need to wait so the `notify()` wrapper can send the result to buildmaster.

nathanjd: Though I think it would be better to just do it inline instead of spawning a new job.

jobs/e2e-test.groovy
michaelmendoza42: nope just weird grammar haha. i rambling a little. i'll clean it up.

jobs/e2e-test.groovy
```diff
@@ -166,176 +191,91 @@ def _setupWebapp() {
...
+      return;
+   }
+
+   // report-merged-results.ts is a new file
+   kaGit.safePullInBranch("webapp/services/static/dev/cypress/e2e/tools", params.CYPRESS_GIT_REVISION);
+
+   dir ('webapp/services/static') {
+      sh("ls ./dev/cypress/e2e/tools");
+      // This script analyzes results and returns the correct error code on failure (1)
+      exec(["npx", "--yes", "tsx", "./dev/cypress/e2e/tools/report-merged-results.ts", *foldersList]);
```
somewhatabstract, line 276: **suggestion[change-requested]:** Based on what LT was doing, it looks like, yes, we wneed to wrap this `catchError` and mark the build result and stage result as unstable when we see test failures.

somewhatabstract (changes requested):
Looking good. Now I see things, I think we need to add the `catchError` stuff as you suggested, but otherwise, I think we're looking good.

somewhatabstract (approved):
I think that some of the job parameters need either deleting or at least marking as:

```
"""IGNORE: This is a dummy parameter that is only here to avoid breaking the
   communication with buildmaster""",
```

For example, SLACK_THREAD, SLACK_CHANNEL, RETRIES...these don't do anything anymore.

somewhatabstract (changes requested):
OOps, didn't mean to approve. Still think the UNSTABLE thing is needed.

jobs/e2e-test.groovy
```diff
@@ -143,199 +147,140 @@ currentBuild.displayName = ("${currentBuild.displayName} " +
                             "(${REVISION_DESCRIPTION})");
 
 // We use the build name as a unique identifier for user notifications.
-BUILD_NAME = "build e2e-cypress-test #${env.BUILD_NUMBER} (${E2E_URL}: ${params.REVISION_DESCRIPTION})"
+BUILD_NAME = "${params.REVISION_DESCRIPTION} ${E2E_URL} #${env.BUILD_NUMBER}"
```
somewhatabstract, line 150: **praise:** nice! This addresses part of FEI-6222

jobs/e2e-test.groovy
```diff
@@ -143,199 +147,140 @@ currentBuild.displayName = ("${currentBuild.displayName} " +
...
-      // deploy command (e.g. `@foo` types: `sun queue foo`).
-      if (n[1] != mainUser) {
-         otherUsers += ",${n[1]}";
+   dir('webapp/services/static') {
+      // start-cy-cloud-run.ts returns non-zero rc if it detects test
+      // failures. We set the job to UNSTABLE in that case because we
+      // report failures later.
+      catchError(buildResult: "UNSTABLE", stageResult: "UNSTABLE",
+            message: "There were test failures!") {
+         exec(runE2ETestsArgs);
```
somewhatabstract, line 244: **question:** I don't think this is right, is it? We want to do the error stuff on the results of running all the tests, not on each agent, right? I think you just need it where you have it for the results, not here.

jobs/e2e-test.groovy
```diff
@@ -143,199 +147,140 @@ currentBuild.displayName = ("${currentBuild.displayName} " +
...
+
+   // report-merged-results.ts is a new file
+   kaGit.safePullInBranch("webapp/services/static/dev/cypress/e2e/tools", params.GIT_REVISION);
+
+   dir ('webapp/services/static') {
+      sh("ls ./dev/cypress/e2e/tools");
+      catchError(buildResult: "UNSTABLE", stageResult: "UNSTABLE",
+            message: "There were test failures!") {
+         exec(["npx", "--yes", "tsx", "./dev/cypress/e2e/tools/report-merged-results.ts", *foldersList]);
+      }
```
somewhatabstract, line 281: **praise:** This looks good to me. I'd like @rgagarinr or @jandrade to confirm but this looks like what we had for LT.

somewhatabstract (commented):
This is closer, but not quite right. I think what makes this harder is that it's hard to see what you're changing because this PR contains both the file rename and the changes. Perhaps split this into two PRs.

1. First, a PR that makes the changes to the e2e-test-cycloud job.
2. Then the PR that merges the two jobs together. 

That way we can more easily see what's changing.

jobs/e2e-test.groovy
michaelmendoza42: ah right. meant to revert this.

jobs/e2e-test.groovy
michaelmendoza42: 🙌

jobs/e2e-test.groovy
michaelmendoza42: 🙌

PR #273, opened 2025-02-20
Prevent e2e-test-cycloud from overwriting smoke test result in buildmaster

nathanjd:
## Summary:
If e2e-test-cycloud finishes after e2e-test, it will overwrite both the
result and job ID.

This has been causing 2 bugs:

1. The link to the e2e job in jenkins points to e2e-test but with the
   e2e-test-cycloud job id. Resulting in a dead link.
2. The status of the smoke test in buildmaster is failed when e2e-test
   passes but e2e-test-cycloud fails.

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1740083704822279?thread_ts=1739930668.893529&cid=C096UP7D0

## Test plan:
Monitor deployment queue.

### DIFF
--- a/jobs/e2e-test-cycloud.groovy
+++ b/jobs/e2e-test-cycloud.groovy
@@ -277,16 +277,18 @@ def analyzeResults(foldersList) {
 }
 
 // Determines if we are running the first or second smoke test.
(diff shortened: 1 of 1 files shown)


michaelmendoza42 (approved):
thanks for figuring this out nathan!

PR #275, opened 2025-02-21
switch to cypress cloud again - this time with hardcoded ka-test-ec2 workers

michaelmendoza42:
## Summary:
hardcode `ka-test-ec2` workers because cypress was complaining about that env was not the same for `ka-firstinqueue-ec2` workers which prevented parallel jobs from running properly

Issue: XXX-XXXX

## Test plan:


jobs/e2e-test.groovy
```diff
@@ -135,240 +125,172 @@ for more information.""",
...
+def analyzeResults(foldersList) {
+   if (currentBuild.result == 'ABORTED') {
+      // No need to report the results in the case of abort!  They will
+      // likely be more confusing than useful.
+      echo('We were aborted; no need to report results.');
+      return;
+   }
+
+   // report-merged-results.ts is a new file
+   kaGit.safePull("webapp/services/static/dev/cypress/e2e/tools");
```
somewhatabstract, line 258: ✅

jobs/e2e-test.groovy
```diff
@@ -135,240 +125,172 @@ for more information.""",
...
-                         what: E2E_RUN_TYPE]]) {
-
-      stage("Sync webapp") {
-         _setupWebapp();
-      }
-
+                   when: ['FAILURE', 'UNSTABLE']]]) {
+       e2e-test.groovy.
+       buildmaster: [sha: params.GIT_REVISION,
+                         what: E2E_RUN_TYPE]
```
somewhatabstract, line 284: ✅

jobs/e2e-test.groovy
```diff
@@ -135,240 +125,172 @@ for more information.""",
...
 
 // At this time removing @ before username.
 DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
 
 // GIT_SHA1 is the sha1 for GIT_REVISION.
 GIT_SHA1 = null;
+REPORT_DIR = "webapp/genfiles"
+REPORT_NAME = "results-combined.json"
 
 // We have a dedicated set of workers for the second smoke test.
```
somewhatabstract, line 151: **suggestion:** Make sure to update this comment to explain why we are ignorring USE_FIRSTINQUEUE_WORKERS

somewhatabstract: Maybe also mark the parameter as ignored

somewhatabstract (approved):
LGTM. A couple of changes I'd like in the commentary (see inline comments).

I presume this also includes the git revision fix we had to make before.

Thanks for your help with all this, @michaelmendoza42; that was a confusing issue until we finally figured it out.

PR #277, opened 2025-02-21
[🔥AUDIT🔥] fix line of code that was commounted out but shouldn't have been

michaelmendoza42:
🖍 _This is an audit!_ 🖍

## Summary:

fix another line that was commented but shouldn't have been

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -275,7 +275,7 @@ def analyzeResults(foldersList) {
 }
 
 // Determines if we are running the first or second smoke test.
-// E2E_RUN_TYPE = (E2E_URL == "https://www.khanacademy.org" ? "second-smoke-test" : "first-smoke-test");
+E2E_RUN_TYPE = (E2E_URL == "https://www.khanacademy.org" ? "second-smoke-test" : "first-smoke-test");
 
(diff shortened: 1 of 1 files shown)


csilvers (commented):
@nathanjd -- you are the one who commented this out originally, in #273.  Was that intentional then, or an oversight?  (It seems unrelated to the fix to me, but I don't know for sure.)

In any case, I'm fine uncommenting this.  If we see problems we can always re-comment it again.

somewhatabstract:
Yup, this is OK. @nathanjd commented this (and some other things) out to prevent the Cypress Cloud job from reporting to buildmaster when it was running side-by-side with the LambdaTest equivalent as that was confusing buildmaster (understandably). After putting Cypress Cloud as our live test runner and having it communicate to buildmaster, we found the issue with 2nd smoke tests and first-in-queue workers. We hadn't identified the problem yet, so we reverted the changes that set things live - which commented the lines out again.

Then @michaelmendoza42 worked on getting Cypress Cloud live again, and that involved reverting the revert, but with additional changes to deal with the first-in-queue workers issue. This specific change was related to getting the Cypress Cloud to report to buildmaster properly where some lines had not been uncommented that should have been.

PR #279, opened 2025-02-24
added more info to the cypress builds

michaelmendoza42:
## Summary:
added more info to the cypress builds

Issue: FEI-6222

## Test plan:
ran script on test job

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -227,7 +227,7 @@ def runE2ETests(workerId) {
    def runE2ETestsArgs = [
       "./dev/cypress/e2e/tools/start-cy-cloud-run.ts",
       "--url=${E2E_URL}",
-      "--name=${BUILD_NAME}",
+      "--name=${BUILD_NAME}"
    ];
 
    dir('webapp/services/static') {
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -10,7 +10,6 @@ import org.khanacademy.Setup;
 //import vars.notify
 //import vars.withTimeout
 
-
 new Setup(steps
 
 ).allowConcurrentBuilds(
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -226,7 +226,7 @@ def runE2ETests(workerId) {
    def runE2ETestsArgs = [
       "./dev/cypress/e2e/tools/start-cy-cloud-run.ts",
       "--url=${E2E_URL}",
-      "--name=${BUILD_NAME}",
+      "--name=${BUILD_NAME}"
```
csilvers, line 229: Why did you delete the comma?  It's good to always put in commas if the language will let you.  (Maybe groovy doesn't?)

jobs/e2e-test.groovy
```diff
@@ -140,7 +140,7 @@ currentBuild.displayName = ("${currentBuild.displayName} " +
    "(${REVISION_DESCRIPTION})");
 
 // We use the build name as a unique identifier for user notifications.
-BUILD_NAME = "${params.REVISION_DESCRIPTION} ${E2E_URL} #${env.BUILD_NUMBER}"
+BUILD_NAME = "${E2E_URL} #${env.BUILD_NUMBER}"
```
csilvers, line 143: Why remove the REVISION_DESCRIPTION from this?  It's useful when looking at tasks via the jenkins UI.

csilvers (approved):
I like adding the extra envvars!  Not sure about the other changes though.

jobs/e2e-test.groovy
michaelmendoza42: oh i thought i left it on accident after removing some debug args. i can put it back. 👍

jobs/e2e-test.groovy
michaelmendoza42: ah ok. we are trying to trim the build name because it gets truncated in the slack message. in that case i can keep it here and use a different variable for what we pass to the slack message 👍

michaelmendoza42: actually looks like we send `params.GIT_REVISION` to the buildmaster so it doesn't come from the build name at all.

PR #280, opened 2025-02-25
Get the build-current-sqlite script running again.

csilvers:
## Summary:
We turned this off when we moved to goliath, because the local
datastore only has user data in it, and that doesn't change very
often.  But it turns out it does change sometimes -- the permissions
granted to testadmin grow over time as more permissions get added to
our system (since testadmin is supposed to have all permissions).  So
there's some value in updating this data periodically.  This new job
does just that.

Issue: https://khanacademy.slack.com/archives/C04SEFXQBNU/p1740171086452009

## Test plan:
I went to
   https://jenkins.khanacademy.org/job/misc/job/build-current-sqlite/1530/replay
and replaced the "Main Script" text with the new script text, except
for the addCronSchedule line, and ran it with success!

### DIFF
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -1,5 +1,7 @@
-// Pipeline job that creates current.sqlite from the sync snapshot
-// (on gcs) and uploads the resulting current.sqlite to gcs.
+// Pipeline job that creates the dev datastore and uploads it to gcs.
+// This is the file that is then downloaded by `make current.sqlite`.
+// (It is also downloaded by the `datstore-emulator` in
+// dev/server/docker-compose.backend.yml.
 
 @Library("kautils")
 // Classes we use, under jenkins-jobs/src/.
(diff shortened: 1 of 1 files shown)


jobs/build-current-sqlite.groovy
```diff
@@ -50,25 +38,28 @@ def runScript() {
...
-   // We need secrets to talk to gcs, prod.
-   withSecrets() {
-      withEnv(
-         ["CURRENT_SQLITE_BUCKET=${params.CURRENT_SQLITE_BUCKET}",
-          "DEV_APPSERVER_ARGS=${params.DEV_APPSERVER_ARGS}",
-          "SNAPSHOT_NAMES=${params.SNAPSHOT_NAMES}"]) {
-         sh("jenkins-jobs/build_current_sqlite.sh");
-      }
+        // First, we need to get a local webserver running.
+        sh("ssh-agent make start-dev-server-backend WORKING_ON=NONE")
```
dbraley, line 42: I think this should be inside the try loop? IIRC, it's possible for the command to fail, but leave containers running? I've only ever noticed it when container startup times out, but still, I know on mac it is possible? Maybe that's an apple oddity that doesn't happen on linux.

dbraley (approved):
This looks fine to me. I suspect the startup should be in the try (as mentioned) and I don't really think our jenkins box is so busy on Sundays we need to start using hash time to plan weekend jobs if I'm being honest, but I don't have a problem with it either (other than maybe it makes it a little tiny bit harder to reason about when the job happens, but meh).

PR #281, opened 2025-02-25
e2e-test minor adjustment to revision description

michaelmendoza42:
## Summary:
made a minor adjustment to the branch info we pass to cypress cloud

Issue: FEI-6222

## Test plan:

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -229,8 +229,6 @@ def runE2ETests(workerId) {
       "./dev/cypress/e2e/tools/start-cy-cloud-run.ts",
       "--url=${E2E_URL}",
       "--name=${BUILD_NAME}",
-      "--targets",
-      "mm-test-flaky"
    ];
 
    dir('webapp/services/static') {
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -133,6 +133,8 @@ NUM_WORKER_MACHINES = null;
 
 // Override the build name by the info that is passed in (from buildmaster).
 REVISION_DESCRIPTION = params.REVISION_DESCRIPTION ?: params.GIT_REVISION;
+// Drop this part so the branches can be grouped in Cypress Cloud
+SHORT_REVISION_DESCRIPTION = REVISION_DESCRIPTION.replaceAll(/\s*\((now live|currently deploying)\)/, '');
```
somewhatabstract, line 137: **question:** Are these the only statuses we append? I had a look and it seems like they are.

jobs/e2e-test.groovy
```diff
@@ -288,7 +290,7 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
       initializeGlobals();
       stage("Run e2e tests") {
          withEnv(["COMMIT_INFO_AUTHOR=${DEPLOYER_USER}",
```
somewhatabstract, line 292: **question[change-requested]:** I think we should drop this too, for now. Let the default commit author fill in instead.

somewhatabstract (approved):
Looks good though we should drop the COMMIT_INFO_AUTHOR setting as part of this too.

PR #282, opened 2025-03-03
Generate skipped list once to avoid google rate limiter

michaelmendoza42:
## Summary:
generate skipped list once
add skip gen flag
unstash skipped list on each worker

Issue: FEI-6282

## Test plan:
tested on debug jenkins job

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -6,9 +6,6 @@
 // Classes we use, under jenkins-jobs/src/.
 import org.khanacademy.Setup;
 // Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
-//import vars.exec
-//import vars.notify
-//import vars.withTimeout
 //import vars.withVirtualenv
 //import vars.exec
 //import vars.notify
@@ -160,8 +157,10 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
 
 // GIT_SHA1 is the sha1 for GIT_REVISION.
 GIT_SHA1 = null;
-REPORT_DIR = "webapp/genfiles"
-REPORT_NAME = "results-combined.json"
+REPORT_DIR = "webapp/genfiles";
+
+SKIPPED_E2E_FILENAME = "skipped_e2e_tests.json";
+SKIPPED_STASH_ID = "e2e-skipped-list";
 
 // We have a dedicated set of workers for the second smoke test. (ka-firstinqueue-ec2)
 // But using them breaks our ability to run e2e in parallel
@@ -243,8 +242,8 @@ def runE2ETests(workerId) {
    ];
 
    dir('webapp/services/static') {
-      unstash "e2e-skipped-list";
-      sh("cat skipped_e2e_tests.json");
+      unstash SKIPPED_STASH_ID;
+      sh("cat ${SKIPPED_E2E_FILENAME}");
       exec(runE2ETestsArgs);
    }
 }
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -9,6 +9,17 @@ import org.khanacademy.Setup;
 //import vars.exec
 //import vars.notify
 //import vars.withTimeout
+//import vars.withVirtualenv
+//import vars.exec
```
csilvers, line 13: exec is already listed on line 9!  Maybe sort and uniqify this list.

jobs/e2e-test.groovy
```diff
@@ -288,8 +301,18 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
            buildmaster: [sha: params.GIT_REVISION,
                          what: E2E_RUN_TYPE]]) {
       initializeGlobals();
+      stage("Generate skipped list") {
+         dir("webapp/services/static") {
+            sh("pnpm cypress:clean");
+            exec(["./dev/cypress/e2e/tools/gen-skipped-e2e-tests.js", "${E2E_URL}"]);
```
csilvers, line 307:
```suggestion
            exec(["./dev/cypress/e2e/tools/gen-skipped-e2e-tests.js", E2E_URL]);
```

jobs/e2e-test.groovy
```diff
@@ -288,8 +301,18 @@ onWorker(WORKER_TYPE, '5h') {     // timeout
...
+            stash includes: "skipped_e2e_tests.json", name: "e2e-skipped-list";
+         };
+      }
       stage("Run e2e tests") {
-         withEnv(["COMMIT_INFO_BRANCH=${SHORT_REVISION_DESCRIPTION}"]) {
+         withEnv([
+            "COMMIT_INFO_BRANCH=${SHORT_REVISION_DESCRIPTION}",
+            "KA_SKIP_GEN=1",
```
csilvers, line 314: Side note: KA_SKIP_GEN is not a very descriptive name.  I had to look at the source to see what it does.  I'd suggest renaming it -- in some other PR -- to KA_DO_NOT_GENERATE_SKIPPED_TESTS or some other long but descriptive name.

csilvers (approved):
Good idea!

jobs/e2e-test.groovy
somewhatabstract: **note:** exec, notify and withTimeout are all duplicated

jobs/e2e-test.groovy
somewhatabstract: `KA_SKIP_GEN` was used to skip various "generation" steps - not just the generation of skipped tests, but also of some other files...I think some have since gone away.

jobs/e2e-test.groovy
```diff
@@ -232,6 +243,8 @@ def runE2ETests(workerId) {
    ];
 
    dir('webapp/services/static') {
+      unstash "e2e-skipped-list";
+      sh("cat skipped_e2e_tests.json");
```
somewhatabstract, line 247: **suggestion:** These two strings (the id to unstash and the filename), are in two places. This feels like it could be easy for them to get out-of-sync; perhaps use variables to store these so that they're only specified in one place?

jobs/e2e-test.groovy
somewhatabstract: No, the files that were skipped before were appropriate in this case too. That's not a concern.

somewhatabstract: The name can definitely be better though; but that's a change for a different PR, as you say. And one that needs to be done carefully since it's cross-repo.

PR #285, opened 2025-03-07
Fix the build-current-sqlite script to actually work.

csilvers:
## Summary:
It had a few problems:

1. It tried to delete the database file after starting the datastore
emulator.  That ain't gonna fly.  Instead, we have to depend on create_dev_users to properly delete old users before creating new ones (which it does).

2. I was using too many `gs://`'s in a row.

3. I was only putting the WEB_INF directory into the tar file, not all
the files that the datastore emulator creates.  This is problematic
because if not all the files are in the tarfile, the datastore
emulator will create them, and they will be owned by root, and we
won't be able to clean them up.

To bootstrap the solution to (3), I manually updated a new version of
dev_datastore.tar to gcs that has all the files in datastore/.  So
next time this script runs, it will populate datastore/ on the build
machine with all the files, and docker won't create any new files
(owned by root), it will just modify the existing files (owned by
jenkins).

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1741306004899129

## Test plan:
Ran this manually via jenkins replay at
   https://jenkins.khanacademy.org/job/misc/job/build-current-sqlite/1541

### DIFF
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -57,16 +57,15 @@ def runScript() {
             sh("ssh-agent env -u BUILD_TAG make start-dev-server-backend WORKING_ON=NONE")
 
             updateDatabase()
-
-            // Now upload the new database to gcs.
-            sh("gsutil cp ${params.CURRENT_SQLITE_BUCKET}/dev_datastore.tar ${params.CURRENT_SQLITE_BUCKET}/dev_datastore.tar.bak")
-            // A rare case we *don't* want the `-t` flag to docker:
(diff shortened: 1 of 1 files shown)


jobs/build-current-sqlite.groovy
```diff
@@ -33,33 +33,42 @@ that haven't yet been merged to master.""",
 ).apply();
 
 
+// This runs in webapp/.
```
MiguelCastillo, line 36: Thanks for this extra comment.  A lot of this stuff has abstractions that makes it not obvious what CWD is.

MiguelCastillo (approved):
Great!

PR #291, opened 2025-03-17
Re-added slack notifications from webapp script

michaelmendoza42:
## Summary:
re-added slack notifications from webapp script
currently, there is no tagged notification on e2e tests passing so deployers have missed the cue if they weren't paying super close attention

Issue: FEI-6290

## Test plan:
ran script on jenkins test job

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -239,6 +239,8 @@ def runE2ETests(workerId) {
       "./dev/cypress/e2e/tools/start-cy-cloud-run.ts",
       "--url=${E2E_URL}",
       "--name=${BUILD_NAME}",
+      "--targets",
+      "mm-test-pass",
    ];
 
    dir('webapp/services/static') {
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -272,18 +272,54 @@ def analyzeResults(foldersList) {
...
-      catchError(buildResult: "UNSTABLE", stageResult: "UNSTABLE",
-         message: "There were test failures!") {
-         exec(["npx", "--yes", "tsx", "./dev/cypress/e2e/tools/report-merged-results.ts", *foldersList]);
+   withTimeout('5m') {
+      // several new files in util and tools
+      kaGit.safePull("webapp/services/static/dev/cypress/e2e/tools");
+      kaGit.safePull("webapp/services/static/dev/cypress/e2e/util");
+
+      dir('webapp/services/static') {
+         // Get the SLACK_TOKEN from global credentials (env vars).
```
csilvers, line 281: Wait, what is this?

We should be getting the slackbot token from gsm, not from jenkins.  Was SLACK_TOKEN something you set up, or something that was already there?

csilvers: @jandrade this looks like something you added?  I think this is something I missed when reviewing #98, on account of that PR being a rename of a file that maybe I didn't review originally.

csilvers (approved):
It looks like this is just reviving code that used to live in e2e-test.groovy before?  I don't really know what this new script does, but I assume it was tested and worked well before.

PR #292, opened 2025-03-17
Allow concurrent builds for delete-versions

jwbron:
## Summary:
I made some improvements to how reaper is triggering delete-versions [here](https://github.com/Khan/buildmaster2/pull/252), but there's still a chance that the queue could get backed up. To help prevent that, let's allow these builds to run concurrently. Reaper triggers separate jobs for each service, so it doesn't seem like there's any reason we should be forcing these to be run sequentially.

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -70,7 +70,7 @@ def deleteVersion() {
 }
 
 
-onWorker('30m') {
+onMaster('30m') {
    notify([slack: [channel: '#1s-and-0s-deploys',
                 sender: 'Mr Monkey',
                 emoji: ':monkey_face:',


alertlib
nathanjd, line 1: Did you intend to delete this?

jobs/delete-version.groovy
```diff
@@ -30,6 +30,10 @@ spaces. Must exactly match the name of an existing version (e.g.
     """The modules to delete, comma-separated.  Must not be empty.""",
     ""
 
+// Buildmaster triggers this job once for each service every 15 minutes.
+// If we disallow concurrent builds, the job queue can get backed up.
+).allowConcurrentBuilds(
```
nathanjd, line 35: Other files have this at the top of the chain. Might be best to keep that consistent.

nathanjd (commented):
It looks good to me but I'd like @MiguelCastillo or @csilvers to review as I don't know why this was forced to be serial on master initially. Seems an odd choice to not be intentional. Maybe the order of deletion is important. Maybe there's some rate limit or GCP limitation.

csilvers (commented):
> It doesn't really make sense to me that we're splitting up [reaper](https://github.com/Khan/buildmaster2/blob/master/src/reaper.py) and [delete_gae_versions](https://github.com/Khan/webapp/blob/master/deploy/delete_gae_versions.py). Why not just mash them together and run this as a single job with no params as opposed to triggering it once for every service that needs to have versions deleted?

We split them up because reaper.py needs access to the buildmaster db to do its work, but the actual version-deletion code (delete_gae_versions.py) depends on a bunch of webapp infrastructure we didn't want to have to re-implement in buildmaster.  To say nothing of getting all the permissions right.

I do like your idea of just passing in a json blob to this job, saying exactly what versions of what services to delete.  (It doesn't need to be json, we could also just do a string like `service:version,version,version service:service,version,version ....` and parse it ourselves.  I don't see any reason not to do that.

I don't see any reason not to run this job concurrently, so I'm approving this change, but please revert the change to use onWorker.  But I'm wondering if we want to just go straight to your pass-in-json instead, and not bother with this.  Since I think it will give the same win.

csilvers (approved):

PR #295, opened 2025-03-18
[🔥AUDIT🔥] Revert "Allow concurrent builds for delete-versions"

jwbron:
🖍 _This is an audit!_ 🖍

## Summary:
Concurrent jobs are hanging. Only the first one is succeeding. Not sure why, but let's revert this and find a better fix.

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -13,10 +13,6 @@ import org.khanacademy.Setup;
 
 new Setup(steps
 
-// Buildmaster triggers this job once for each service every 15 minutes.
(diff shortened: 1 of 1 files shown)


csilvers (commented):
Ok. Can you point me to a hanging job?

csilvers:
Oh, it's not hanging.  It's just creating a new workspace and needs to clone webapp into that workspace, which takes forever.  You just needed to wait.

csilvers:
When you run concurrent jobs, you need a lot of workspaces for them.  It's a one-time cost.

PR #296, opened 2025-03-18
Allow versions for multiple services to be deleted in one job

jwbron:
## Summary:
See https://github.com/Khan/buildmaster2/pull/253

We're having some issues with the queue on master getting a bit backlogged and it seems one of the causes is reaper. Every 15 minutes, reaper spins up several jobs (that only take a minute or so to run), but since the jobs are run sequentially, a backlog can build up if there's any slowness from jenkins. To remediate that, let's just trigger a single job that handles all services.

Issue: INFRA-XXXX

## Test plan:
Try running this job manually in jenkins using a replay

### DIFF
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -36,7 +36,7 @@ def verifyArgs() {
       notify.fail("The SERVICE_VERSIONS parameter is required.");
    }
    try {
-      def parsed = new JsonSlurper().parseText(params.SERVICE_VERSIONS);
+      def parsed = new JsonSlurperClassic().parseText(params.SERVICE_VERSIONS);
       if (!(parsed instanceof Map)) {
          notify.fail("SERVICE_VERSIONS must be a JSON object.");
       }


jobs/delete-version.groovy
```diff
@@ -51,21 +64,18 @@ def _setupWebapp() {
...
-         if (params.MODULES) {
-            args += ["--modules",
-                     params.MODULES.split(/,/).collect { it.trim() }.join(',')
-                     ];
+         def serviceVersions = new JsonSlurper().parseText(params.SERVICE_VERSIONS);
+
+         serviceVersions.each { service, versions ->
+            def args = ["deploy/delete_gae_versions.py"]
+            args += (versions as List)  // Ensure Groovy treats it as a list
+            args += ["--modules", service]
```
nathanjd, line 72: Do we still want to trim each service name?

nathanjd (approved):
Seems like a good improvement to me.

csilvers (changes requested):
Wow, you are showing some real groovy-fu!

As I mentioned in https://github.com/Khan/buildmaster2/pull/253#issuecomment-2737488593, I think you want to support both invocation methods here for now.

I don't think that's difficult.  You can update verifyArgs to check if either SERVICE_VERISONS or VERSIONS is set, and complain if neither is, and then sanity-check SERVICE_VERSIONS like you're doing today.  You could also check, if you wanted, that only one of VERSION and SERVICE_VERSIONS is set, and if SERVICE_VERSIONS is set, MODULES is not.  But that's probably overkill; we're not trying to protect against malicious users here.

And then in deleteVersion(), you can do something lke:
```
args = ["deploy/delete_gae_versions.py"];
if (parmas.VERSION) {
    args += ...;
}
if (params.MODULES) {
   args += ...;
}
if (params.SERVICE_VERSIONS) {
   args += ...;
}
```

csilvers (approved):
OK, I'll approve based on your plan from https://github.com/Khan/buildmaster2/pull/253#issuecomment-2737507755.  But please fix up the semicolons first!

PR #297, opened 2025-03-19
[FEI-6189] Disallow deploying frontend and backend changes at the same time.

jeresig:
## Summary:
We want to disallow deploying frontend and backend changes together (in general and in preperation for moving the frontend code out into a new repo).

We're not going to deploy this until we're absolutely sure that we've figured out any edge cases, but it's likely that we're really close to deploying this.

All the changes we've made are in these two epics:
https://khanacademy.atlassian.net/browse/FEI-6194
https://khanacademy.atlassian.net/browse/FEI-6178

Issue: FEI-6189

## Test plan:
Not sure how to test this... I think this is syntactically correct, but I assume the test plan is probably deploying this and then confirming that a plain static (or non-static) deploy continues to work - and that a combination deploy correctly fails.

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -344,6 +344,17 @@ def initializeGlobals() {
             SERVICES = [];
          }
 
+         // If we're deploying static and other services at the same time,
+         // we want to disallow this as we're going to be moving the static
+         // service out of webapp. It can be overridden with the FORCE flag.
+         if ("static" in SERVICES && SERVICES.size() > 1 && !params.FORCE) {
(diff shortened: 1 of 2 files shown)


jeresig:
@jwbron I'm not sure that we know which services a branch may deploy until we run `deploy/should_deploy.py` - and the first time we do it is just above the statements that I added. I don't think we know until we do the branch merge and run this script (at which point it'll fail and not do any builds).

somewhatabstract (approved):
LGTM. Excited to see this getting enforced more directly. Really great to see that we have an escape hatch too.

PR #299, opened 2025-03-19
Post to `#whats-happening` when we deploy to fastly.

csilvers:
## Summary:
We have a goal of posting to the `#whats-happening` slack channel
whenever we deploy a change to our production system.  Currently,
changes to our fastly configs are noticed, via polling, by a cronjob
running on toby.  But now that all fastly deploys are (supposed to) go
through this jenkins job, it can do the reporting to #whats-happening
instead!

Once this is live, I'll disable the toby cronjob.  Or rather, I'll
repurpose it to notice when deploys to fastly are made via the UI and
not via the fastly deploy process.

Issue: https://khanacademy.slack.com/archives/C06T4JE5HCN/p1741644811156629

## Test plan:
Fingers crossed, we'll see what happens next time we need to deploy
prod vcl!

### DIFF
--- a/jobs/deploy-fastly.groovy
+++ b/jobs/deploy-fastly.groovy
@@ -169,9 +169,9 @@ def notifyWithVersionInfo(oldActive, newActive) {
    // automatically as part of the normal deploy process, and gets
    // alerted on #whats-happening via that mechanism.
    if (params.TARGET == "prod" && params.SERVICE != "khanacademy-org-compute") {
-      def subject = "Fastly service ${params.SERVICE} was modified.";
-      def body = "New version: ${newActive}";
-      def cmd = [
+      subject = "Fastly service ${params.SERVICE} was modified.";
+      body = "New version: ${newActive}";
+      cmd = [
           "jenkins-jobs/alertlib/alert.py",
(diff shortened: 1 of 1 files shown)


nathanjd (approved):

jobs/deploy-fastly.groovy
```diff
@@ -163,6 +163,35 @@ def notifyWithVersionInfo(oldActive, newActive) {
    withSecrets.slackAlertlibOnly() {
       sh("echo ${exec.shellEscape(body)} | ${exec.shellEscapeList(cmd)}");
    }
+
+   // Let's tell #whats-happening as well, at least for prod deploys.
+   // But we don't alert for fastly-compute, since that is deployed
+   // automatically as part of the normal deploy process, and gets
+   // alerted on #whats-happening via that mechanism.
+   if (params.TARGET == "prod" && params.SERVICE != "khanacademy-org-compute") {
```
MiguelCastillo, line 171: I was initially thinking that we'd do it for just VCL. But doing it for all of them seems useful!

MiguelCastillo (approved):
shippit!

PR #300, opened 2025-03-20
Avoid passing in out of date param to display name

jwbron:
## Summary:
I missed a reference to an old param. This isn't causing any failures, but all jobs have `(null)` in the display name now

Issue: INFRA-XXXX

## Test plan:

### DIFF
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -28,6 +28,9 @@ Example: {"serviceA": ["version1", "version2"], "serviceB": ["version3"]}""",
 
 ).apply();
 
+currentBuild.displayName = "${currentBuild.displayName} (${params.SERVICE_VERSIONS})";
+
+
 def verifyArgs() {
    if (!params.SERVICE_VERSIONS) {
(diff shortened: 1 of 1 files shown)


jobs/delete-version.groovy
```diff
@@ -28,9 +28,6 @@ Example: {"serviceA": ["version1", "version2"], "serviceB": ["version3"]}""",
 
 ).apply();
 
-currentBuild.displayName = "${currentBuild.displayName} (${params.VERSION})";
```
csilvers, line 31: How would it look to change this to:
```
currentBuild.displayName = "${currentBuild.displayName} (${params.SERVICE_VERSIONS})";
```
?

I know it's really wordy, but I wonder if it's unusably wordy.  I do say that the old code (that showed the version) was useful to me from time to time to figure out what has been reaped and what hasn't.

csilvers (approved):
I'm ok removing this entirely, but if we could show the params in there it would be even more useful, to me!

nathanjd (approved):

PR #302, opened 2025-03-20
Notify when someone manually deploys a change to fastly.

csilvers:
## Summary:
All of our fastly services are intended to be deployed via the
fastly-deploy jenkins job.  However, sometimes people deploy a fastly
service manually, by going to the fastly UI, hitting "clone", making
some changes, and hitting "save".  We do not want such changes to
happen, since there is no record of them in source control, and indeed
the change will be overwritten the next time a "proper" deploy
happens, via fastly-deploy.

This new script notices when such deploys happen.  It works in concert
with the fastly-deploy job to do its work: fastly-deploy says what
versions it has deployed, and we talk to the fastly API to find out
what versions are live that were *not* deployed by fastly-deploy.
When we see one, we alert in an alerting channel, and also in
production system.

A similar script used to exist on toby.  This PR moves it here, now
that it need to work in concert with the fastly-deploy jenkins job.
Once this is working I'll remove the old script from toby.

Issue: https://khanacademy.slack.com/archives/C06T4JE5HCN/p1741644811156629

## Test plan:
I ran this manually on jenkins.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.egg-info
+__pycache__
--- a/fastly_notifier.py
+++ b/fastly_notifier.py
@@ -175,7 +175,9 @@ if __name__ == '__main__':
         with open(args.bad_deploys_file) as f:
             bad_deploys_by_service_id = _parse_deploys_file(f)
     else:
-        logging.warning("No history file found at %s", args.bad_deploys_file)
+        # It's expected this will be empty if nobody has ever done
+        # anything wrong!
+        logging.debug("No history file found at %s", args.bad_deploys_file)
         bad_deploys_by_service_id = {}
 
     service_info = get_service_info(api_key)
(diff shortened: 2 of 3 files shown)


MiguelCastillo (approved):
This very lil code but it is very well thought out.  Thank you!

At some point we were thinking about updating https://github.com/Khan/internal-webserver/blob/master/fastly_notifier.py.

Should we do that?  I'd like to be able to tell when this happens before the next deploy, which sometimes VCL isn't deployed for some time.

csilvers:
> Should we do that? I'd like to be able to tell when this happens before the next deploy, which sometimes VCL isn't deployed for some time.

Actually, nothing runs this script at the moment.  I plan on adding it to a cronjob, but I have to do that in a separate repo.

csilvers:
I was going to put it in Khan/aws-config:jenkins/crontab

MiguelCastillo:
> I was going to put it in Khan/aws-config:jenkins/crontab

Oh right, I see.  That makes sense.  thank you for clarifying.

PR #305, opened 2025-03-27
Compute services to deploy during merge-branches

nathanjd:
## Summary:
To simplify the logic for automated aborts when doing an unattended
deploy, we should tell buildmaster what services a deploy should be
deploying during merge-branches instead of during build-webapp. This is
so failures during build-webapp or webapp-test (run in parallel) always
have access to the list of services and therefore the computed
attendedness value.

This change will happen in 5 stages:

1. Deploy this PR to declare the new parameters for merge-branches. This
   will also allow the services output to be checked in the logs.
   merge-branches will not send services to buildmaster at this time.
2. Update buildmaster to pass in the new parameters params.SERVICES and
   params.BASE_REVISION to merge-branches and accept services from
   PATCH /commits/merge.
3. Update merge-branches to send the list of services to buildmaster. At
   this time, build-webapp should overwrite the same services.
4. Update buildmaster to send the list of services received from 
    merge-branches to build-webapp so it doesn't need to recompute them.
5. Update build-webapp to no longer run webapp/deploy/should_deploy.py.

4 and 5 don't need to be done immediately, but should be done soon to
speed deployments up.

Unfortunately, adding the webapp/deploy/should_deploy.py call to
merge-branches adds ~40 seconds to its execution time. Bringing the
total from ~5 seconds to ~45 seconds. This is increased to 90 seconds if
the virtual environment must built from scratch. Hopefully, we make this
time back in aggregate by removing it from build-webapp and being able
to dequeue unattended builds sooner.

An example replay execution of merge-branches with this code is here:

https://jenkins.khanacademy.org/job/deploy/job/merge-branches/298367/

Issue: https://khanacademy.atlassian.net/browse/INFRA-10586

Test plan:

1. Deploy to buildmaster prod.
2. Queue a static-only deploy in #1s-and-0s-deploys.
3. Verify merge-branches was successful.
4. Verify in the merge-branches logs that the services were correctly
   detected as "static".
5. Dequeue

### DIFF
--- a/jobs/merge-branches.groovy
+++ b/jobs/merge-branches.groovy
@@ -26,6 +26,13 @@ new Setup(steps
    ""
 
 ).addStringParam(
+   "BASE_REVISION",
+   """Compute services that would be deployed by the merged commit since this 
+   revision based on the the commits between BASE_REVISION..MERGED_SHA1. 
+   This only matters if params.SERVICES is 'auto'.""",
+   ""
+
+).addStringParam(
    "COMMIT_ID",
    """<b>REQUIRED</b>. The buildmaster's commit ID for the commit we will
 create.""",
@@ -60,6 +67,15 @@ of the GIT_REVISION, especially if it is a commit rather than a branch.""",
    ""
 
 ).addStringParam(
+   "SERVICES",
+   """A comma-separated list of services this commit would deploy, or the  
+   special value 'auto', which says to choose the services to deploy 
+   automatically based on what files have changed.  For example, you might 
+   specify "users,static" to force a full deploy to the users service and GCS.
+   """,
+   "auto"
+
+).addStringParam(
    "BUILDMASTER_DEPLOY_ID",
    """Set by the buildmaster, can be used by scripts to associate jobs
 that are part of the same deploy.  Write-only; not used by this script.""",
@@ -68,7 +84,6 @@ that are part of the same deploy.  Write-only; not used by this script.""",
 ).apply();
 
 currentBuild.displayName = "${currentBuild.displayName} (${params.COMMIT_ID}: ${params.GIT_REVISIONS}) (${params.REVISION_DESCRIPTION})";
-
 
 def checkArgs() {
    if (!params.GIT_REVISIONS) {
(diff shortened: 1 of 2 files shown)


jobs/merge-branches.groovy
```diff
@@ -79,15 +94,58 @@ def checkArgs() {
 }
 
 
-def getGaeVersionName() {
+String getGaeVersionName() {
```
csilvers, line 97: This may be a good idea, but it's really different from all the other jenkins jobs we have.  I'd just use `def` everywhere.  (Or change everyone to use the return type in all our jobs, but I don't think you want to take that on.)

csilvers (commented):
This looks correct, I assume you just copied the relevant groovy code from build-webapp.groovy?  But we will have to decide if we want to deploy this, as a matter of principle.

One thing we could do is only run it when SERVICES is set (and not have it default to `auto`).  And have buildmaster only pass it in when we're nearing the top of the queue.  And then maybe change build-webapp, likewise, to only compute should-deploy when SERVICES is set.  And buildmaster will pass in SERVICES to build-webapp if it didn't already git it from merge-branches for some reason.  But that's getting pretty complicated and I don't know how I feel about it.

nathanjd:
> OK, I am remembering why we do should_deploy in build-webapp instead of merge-branches: it just runs a lot less often. (build-webapp only runs once you're in the top 3 on the queue). Our current jenkins job-id for merge-branches is 298575, while build-webapp is at 61226. Especially given how slow should-deploy takes to run, I wonder if want to be running it 5 times as often.

Unfortunately, while we only run build-webapp for the front 3 of the queue, we run webapp-test for all positions in queue which would need this information upon failure.

I suppose one alternative would be to have a dedicated job for services that is only run when we either need to build or in the event of a test failure.

```
merge-branches  -> test
                -> services (if 3rd or earlier in queue or test failure) -> build
```

That way, if there was a test failure, we wouldn't dequeue immediately but would dequeue ~1 minute later after the services job reports back. Thoughts?

nathanjd:
> What if we add a flag to merge-branches to calculate the services being deployed, but we don't use it by default. But buildmaster, if it ever needs to know the services for whatever reason, will call that job and then wait for the data to come back before continuing?
>
> That's adding more states to buildmaster, which maybe isn't a good idea.

Hmm. I don't like that, in the event a commit progresses to 3rd in queue, we would need to re-run the merge-branches job (this time with the flag). I think the states for that would get messier than if we had a dedicated services job which we invoked in a similar pattern.

csilvers:
No, you don't run it with the flag by default.  Only if a test fails and you don't have the service-information at that time.

It doesn't need to be part of merge-branches, it could be a separate job.  Maybe that would make more sense in any case.

nathanjd:
> No, you don't run it with the flag by default. Only if a test fails and you don't have the service-information at that time.

Right, that's what I was trying to say. I think we're in agreement.

> It doesn't need to be part of merge-branches, it could be a separate job. Maybe that would make more sense in any case.

I do think a separate job is the way to go if we want to avoid wasted work. I'll come back with a PR for that.

nathanjd:
New PR

https://github.com/Khan/jenkins-jobs/pull/306

PR #306, opened 2025-03-29
Add determine-webapp-services job

nathanjd:
## Summary:
To simplify the logic for automated dequeues when doing an unattended
deploy, we should tell buildmaster what services a deploy should be
deploying before starting build-webapp. This is so failures during
build-webapp always have access to the list of services and therefore
the computed attendedness value.

Failures in webapp-test may need to trigger this job early in order to
tell if it should dequeue.

This job should not be started if the deployer has queued with the
SERVICES flag because that flag would override the result of this job.

This change will happen in 4 stages:

1. Deploy this PR to to perform testing with manual builds.
2. Update buildmaster to represent and call this job with a new state
   machine. At this time, build-webapp should overwrite the same
   services.
3. Update buildmaster to send the list of services received from this
   job to build-webapp so it doesn't need to recompute them.
4. Update build-webapp to no longer run webapp/deploy/should_deploy.py.

Issue: https://khanacademy.atlassian.net/browse/INFRA-10586

## Test plan:
1. Deploy jenkins job.
2. Manually start builds for the job.
3. Verify the list of services is output correctly into the logs.

### DIFF
--- a/jobs/determine-webapp-services.groovy
+++ b/jobs/determine-webapp-services.groovy
@@ -1,0 +1,162 @@
+// Determine what services would be deployed by a webapp commit. Should not
+// be started if the deployer has queued with the SERVICES flag because that
+// flag would override the result of this job.
+
+@Library("kautils")
+// Classes we use, under jenkins-jobs/src/.
+import org.khanacademy.Setup;
+// Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
+//import vars.buildmaster
+//import vars.exec
+//import vars.kaGit
+//import vars.notify
+//import vars.withTimeout
+
+new Setup(steps
+
+).allowConcurrentBuilds(
+
+).addStringParam(
+    "GIT_REVISION",
+    """<b>REQUIRED</b>. The sha1 to determine the deployed services for.""",
+    ""
+
+).addStringParam(
+   "BASE_REVISION",
+   """Compute services that would be deployed by the commit since this 
+   revision based on the the commits between BASE_REVISION..GIT_REVISION.""",
+   ""
+
+).addStringParam(
+   "DEPLOYER_USERNAME",
+   """Who asked to run this job, used to ping on slack.
+If not specified, guess from the username of the person who started
+this job in Jenkins.  Typically not set manually, but by hubot scripts
+such as sun.  You can, but need not, include the leading `@`.""",
+   ""
+
+).addStringParam(
+   "SLACK_CHANNEL",
+   "The slack channel to which to send failure alerts.",
+   "#1s-and-0s-deploys"
+
+).addStringParam(
+   "SLACK_THREAD",
+   """The slack thread (must be in SLACK_CHANNEL) to which to send failure
+alerts.  By default we do not send in a thread.  Generally only set by the
+buildmaster, to the 'thread_ts' or 'timestamp' value returned by the Slack
+API.""", 
+   ""
+
+).addStringParam(
+   "REVISION_DESCRIPTION",
+   """Set by the buildmaster to give a more human-readable description
+of the GIT_REVISION, especially if it is a commit rather than a branch.""",
+   ""
+
+).addStringParam(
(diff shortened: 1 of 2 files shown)


jobs/determine-webapp-services.groovy
```diff
@@ -0,0 +1,162 @@
...
+   if (!params.GIT_REVISION) {
+      notify.fail('The GIT_REVISION parameter is required.');
+   }
+}
+
+def checkoutWebapp() {
+   deployer_username = notify.getDeployerUsername(params.DEPLOYER_USERNAME)
+    
+   // TODO(csilvers): have these return an error message instead of alerting 
+   // themselves, so we can use notify.fail().
```
csilvers, line 88: Copypasta?

jobs/determine-webapp-services.groovy
```diff
@@ -0,0 +1,162 @@
...
+   withEnv(["SLACK_CHANNEL=${params.SLACK_CHANNEL}",
+            "SLACK_THREAD=${params.SLACK_THREAD}",
+            "DEPLOYER_USERNAME=${deployer_username}",
+            "JOB_PRIORITY=${params.JOB_PRIORITY}"]) {
+      kaGit.safeSyncToOrigin("git@github.com:Khan/webapp",
+                             params.GIT_REVISION)
+   }
+}
+
+String[] determineServicesToDeploy() {
```
csilvers, line 98: Please use `def` just to be consistent with other code in this repo.  And every place you use `String` below.

vars/notify.groovy
```diff
@@ -133,6 +133,26 @@ def _sendToAlertlib(subject, severity, body, extraFlags) {
 }
 
 
+// Return the deployer username, ensuring it begins with @ and falling back to
+// the jenkins build user ID if no username is passed.
+String getDeployerUsername(String deployerUsernameParam) {
```
csilvers, line 138: Do you want to change build-webapp.groovy to call this?  Seems a shame to have this in a shared location and not use it twice!

I also wonder if deploy-webapp.groovy should use it, but the code there is different and I'm not sure why.

csilvers (approved):
Looks straightforward enough!  Just go Old Skool with the `def` everywhere and it's good to check in.

vars/notify.groovy
nathanjd: Yeah I do want to use this in other locations but I wanted to make sure it was working here first. Can't do a replay test of a job that doesn't exist yet!

jobs/determine-webapp-services.groovy
nathanjd: Yes, I just copied the code from build-webapp, which I plan to delete so I thought it best to preserve the TODO.

vars/notify.groovy
csilvers: Sounds good, I'd suggest putting in a TODO in the places you want to use it.

jobs/determine-webapp-services.groovy
nathanjd: I was thinking of doing this piecemeal since typing is optional but may end up throwing errors during runtime. Changing the whole repo at once seems high risk. I'll remove the typing for now, as requested.

csilvers: I'm on board with doing it piecemeal!, but let's do it a job at a time, rather than mixing and matching within a single file.

nathanjd: Ahh, well then what do you think about this being the first file? Some of the function definitions are still `def` which is  correct in the typing because they don't have a return type.

https://groovy-lang.org/objectorientation.html#_method_definition

nathanjd: I've tried researching and using the groovy playground to see if it's possible to have a return type of null, for example: `Null functionName(){}`. I was not able to get syntax that didn't produce a compilation error. Groovy also implicitly returns the result of the last line executed in a function so I think the "typed function that returns nothing" should be:

```
def foo(){
    // do things
    return null
}
```

PR #307, opened 2025-03-31
Add environment variable for jenkins-deploy service account auth to jenkins server

jwbron:
## Summary:
Apparently [jenkins jobs don't source ~/.profile](https://github.com/Khan/aws-config/pull/145#discussion_r2021785324), so the `GOOGLE_APPLICATION_CREDENTIALS` env var needs to be defined at runtime. This env var points to the auth json used for gcp authentication by gcloud and gcp SDKs. It's necessary for the switchover from the prod-deploy user account to the jenkins-deploy service account.

Issue: INFRA-10547

## Test plan:
n/a


csilvers (approved):
I don't know if it's safe to set onWorker before we actually change the worker to have the service account installed; I don't know exactly how the envvar is used.  Let's just change onMaster now, and onWorker later.  Once that's done, shipit

PR #308, opened 2025-03-31
Add environment variable for jenkins-deploy service account auth on workers

jwbron:
## Summary:
Apparently [jenkins jobs don't source ~/.profile](https://github.com/Khan/aws-config/pull/145#discussion_r2021785324), so the `GOOGLE_APPLICATION_CREDENTIALS` env var needs to be defined at runtime. This env var points to the auth json used for gcp authentication by gcloud and gcp SDKs. It's necessary for the switchover from the prod-deploy user account to the jenkins-deploy service account.

Issue: INFRA-10547

## Test plan:
n/a


nathanjd (approved):
I guess an alternative would be to source .profile but that is a big change.

csilvers (approved):
Approving for when the time comes!

PR #309, opened 2025-03-31
Revert env changes in #307

jwbron:
## Summary:
We're seeing some issues caused by the env var set in #307, so let's revert it for now

Issue: INFRA-XXXX

## Test plan:


csilvers (approved):
shipit!

PR #312, opened 2025-04-07
[🔥AUDIT🔥] Fix the service-dir for fastly-vcl.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
I've moved the directory to its standard location inside webapp, so we
don't need to special-case it in this jenkins job anymore!

Issue: https://khanacademy.atlassian.net/browse/INFRA-10538

## Test plan:
Fingers crossed


MiguelCastillo (commented):
Oh good point.

PR #314, opened 2025-04-12
[🔥AUDIT🔥] Alert to #whats-happening like we said we were going to, on fastly deploys.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
I don't know why I was sending the message to #fastly instead.  That
channel already gets plenty of updates while a fastly deploy is
happening.

Issue: https://khanacademy.slack.com/archives/CCCC6HZ4K/p1744413054178629

## Test plan:
Fingers crossed


MiguelCastillo (commented):
thanks!

PR #315, opened 2025-04-14
Decrease build-webapp timeout to 1h

jwbron:
## Summary:
[See here for context](https://khanacademy.slack.com/archives/C096UP7D0/p1744668267796409). build-webapp keeps hanging, presumably due to memory issues. Let's decrease the timeout to catch unattended deploys a bit sooner
Issue: INFRA-XXXX

## Test plan:

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -922,7 +922,7 @@ def finishWithFailure(why) {
 
 // We do promotes on master, to ease debugging and such.  Promote isn't
 // CPU-bound, and we can have only one at a time, so it's not a problem.
-onMaster('4h') {
+onMaster('1h') {
    notify([slack: [channel: '#1s-and-0s-deploys',
                    sender: 'Mr Monkey',
                    emoji: ':monkey_face:',


jeresig (approved):

jobs/deploy-webapp.groovy
```diff
@@ -922,7 +922,7 @@ def finishWithFailure(why) {
 
 // We do promotes on master, to ease debugging and such.  Promote isn't
 // CPU-bound, and we can have only one at a time, so it's not a problem.
-onMaster('4h') {
+onMaster('1h') {
```
nathanjd, line 925: If we're dropping this to 1h. We should reduce the two 1h timeouts for mergeFromMaster() and initializeGlobals().

nathanjd: O there's more.

stage("Deploying") has a 150m timeout

finishWithFailure() has a 20m timeout which seems super excessive for just sending some alerts.

nathanjd: I don't think this is so time sensitive we need to be rushing updates out and then cleaning up the code later. I don't think we need to do any changes beyond the 15 minute timeout in `deployToGCS()` and I would prefer the code make sense. Having nested timeouts that are larger than their parent is confusing.

nathanjd: It's not like we can test this on a tight feedback loop anyway. The hanging builds are quite intermittent.

PR #316, opened 2025-04-14
Set 15 minute timeout for static build and deploy

nathanjd:
## Summary:
We are seeing rspack hang at this point:

```
13:47:58  > [3/4] Optimize modules
13:47:58  	[70%] sealing - finish modules
```

https://jenkins.khanacademy.org/job/deploy/job/build-webapp/62204/console

In previous investigations, this was determined to be caused by running
out of memory. Bumping the memory for the node process fixed this for
local dev. We tried bumping this even higher for jenkins but saw little
to no reduction in intermittent hanging builds.

https://github.com/Khan/webapp/pull/31069

When this process hangs, there is no reported failure for jenkins so
this often causes deploys to block the deploy queue without any signal
that it is misbehaving. Let's add a 15 minute timeout so the build will
actually report failure instead of just hanging.

https://khanacademy.slack.com/archives/C096UP7D0/p1744670553934739?thread_ts=1744668267.796409&cid=C096UP7D0

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1744668267796409

## Test plan:
No guarunteed repro case. Deploy and monitor build-webapp job.

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -386,24 +386,29 @@ def deployToGCS() {
    if (!("static" in SERVICES)) {
       return;
    }
-   def args = ["deploy/deploy_to_gcs.py", NEW_VERSION,
-               "--slack-channel=${params.SLACK_CHANNEL}",
-               "--deployer-username=${DEPLOYER_USERNAME}",
-               // We don't send the changelog in a build-only context:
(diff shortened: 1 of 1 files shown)


jeresig (approved):
Great, thank you!

PR #318, opened 2025-04-22
fix(update-ownership-data,update-devserver-static-images): remove success messages from slack alerts to reduce slack noise

zaquestion:
## Summary:
These success messages account for nearly 14% of the messages in #infrastructure for the last year. I think we only really care about when they fail and that they return to normal. Ive adjusted the when conditions to mimic the same behavior as "failureChannel" and removed failureChannel where it was redundant

Issue: https://khanacademy.atlassian.net/wiki/spaces/~712020c9d02fe34fc6401588cedd4b2aaa0680/pages/3960045676/2025+Alerts+Audit

## Test plan:
- look for success messages to be removed from slack

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -39,7 +39,7 @@ the updated data-file to it.""",
 ).addStringParam(
    "SLACK_CHANNEL",
    "The slack channel to send our status info.",
-   "#infrastructure"
+   "#local-devserver"
 
 ).addCronSchedule(
    '0 22 * * *'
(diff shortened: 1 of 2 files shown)


csilvers (commented):
Heh, those are the only two messages on #infrastructure that I actually find useful!  I like seeing the daily update that it's successful, because sometimes the failure is in the jenkins job itself, such that it doesn't send out any message at all.  So we can't depend on "no news is good news," like your PR would make things.

Rather than the raw number of messages, I think what is problematic from my point of view is the bursty messages.  Like 10 messages come in at the same time about this and that route being slow.  It's overwhelming in a way that these two messages aren't.

PR #321, opened 2025-04-25
Update the graphql safelist when we deploy znds

MiguelCastillo:
## Summary:
We are having an issue where updates to queries in from manually created znds are failing to pass our safelisting checks in the graphql gateway, which causes requests from znds to fail.  This regressed when we updated how we check for permissions of safelisted queries some time ago.

So in this PR I am making sure that we are adding queries from znds that engineers create themselves via jenkins. What Im not doing however, is priming queryplanner caches since these types of znds usually don't need to pre create all the queryplans as engineers tend to use a small subset of the UI.

I found this when I was testing adding the x-trace header to tasks we queue from our backends, which can be seen here https://cloudlogging.app.goo.gl/sHnpaBj8Fke8myCp7.

Issue: INFRA-10640 https://github.com/Khan/webapp/pull/31675#issuecomment-2828916672

## Test plan:
Once these changes are done, deploy a znd with a changed opname and verify that works.

### DIFF
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -272,7 +272,7 @@ def uploadGraphqlSafelist() {
       dir("webapp") {
          exec([
             "deploy/upload_graphql_safelist.py",
-            NEW_VERSION,
+            VERSION,
             "--prod",
          ])
       }
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -256,6 +256,29 @@ def deployToGatewayConfig() {
    }
 }
 
+// Updates the safelist so that znds with changes to queries can make our
+// secutiry/permissions checks happy.
+//
+// We do not however, prime the queryplan caches for znds.  Most often folks
+// creating znds need to access a small number of queries so queryplan priming
+// can be left as a dynamic thing the graphql gateway does for znds.
+def uploadGraphqlSafelist() {
+   // We don't upload queries from the static service here becuase
+   // services/static/deploy/deploy.js is responsible for that.
+   // TODO(kevinb): update deploy scripts for each service to be responsible
+   // for uploading its own queries to the safelist.
+   if (SERVICES.any { it != 'static'}) {
+      echo("Uploading GraphQL queries to the safelist.");
+      dir("webapp") {
+         exec([
+            "deploy/upload_graphql_safelist.py",
+            NEW_VERSION,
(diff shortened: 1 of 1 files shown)


jobs/deploy-znd.groovy
```diff
@@ -256,6 +256,29 @@ def deployToGatewayConfig() {
...
+   // We don't upload queries from the static service here becuase
+   // services/static/deploy/deploy.js is responsible for that.
+   // TODO(kevinb): update deploy scripts for each service to be responsible
+   // for uploading its own queries to the safelist.
+   if (SERVICES.any { it != 'static'}) {
+      echo("Uploading GraphQL queries to the safelist.");
+      dir("webapp") {
+         exec([
+            "deploy/upload_graphql_safelist.py",
+            NEW_VERSION,
```
MiguelCastillo, line 275: need to fix this.

nathanjd, line 275: `NEW_VERSION` doesn't look defined to me.

MiguelCastillo: yup - are right.  the correct value here is `VERSION`.  PR has been updated

nathanjd (approved):
Makes sense to me!

csilvers (approved):
Gosh, I knew we uploaded the new graphql schema for znd's, but it never occurred to me we were missing the safelist!  Oops.  How did this not bite us like a zillion times in the past?  Maybe it did.

csilvers:
One thing to think about -- this will mean we end up with a bunch more unused queries in the safelist (for znd's that get further modified before landing in prod).  They should expire after a month, but I wonder if we want to change services/graphql-gateway/safelist/prune.go to make the expiry for znd-only queries be sooner than a month.  That's just an optimization though.

MiguelCastillo:
> One thing to think about -- this will mean we end up with a bunch more unused queries in the safelist (for znd's that get further modified before landing in prod). They should expire after a month, but I wonder if we want to change services/graphql-gateway/safelist/prune.go to make the expiry for znd-only queries be sooner than a month. That's just an optimization though.

You know - I was thinking about that too!  Usually folks wont have znds for more than a few days, so it makes sense if we wanted to be more aggressive about deleting them.  But we will want to have a marker so that if those changes endup being deployed as is, we don't delete something that is safelisted as a znd and endup getting deleted.  Just an implementation detail tho.  Perhaps we use the `client_source`...

PR #325, opened 2025-05-22
Don't notify on fastly-deploys that have just happened.

csilvers:
## Summary:
We saw an alert in slack that a fastly deploy was happening outside of
jenkins, but the deploy was actually happening within jenkins.  The
issue was that the notifier ran in the span between when jenkins
finished deploying and jenkins updated the to-ignore file to include
the new deploy.

There are various ways to avoid this kind of false positive.  One is
to use a lock when deploying, but that turned out to be a fair bit of
work to share the lock between the deploy groovy-code and this script.
Easier is to just tell this script to ignore very recent deploys, on
the theory they might still be deploying.  If they are really
non-jenkins deploys, we'll catch that fact on the next run of this
cronjob.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1747772763424209

## Test plan:
Fingers crossed!

Subscribers: @dnerdy

### DIFF
--- a/fastly_notifier.py
+++ b/fastly_notifier.py
@@ -21,6 +21,7 @@ production system.
 This is meant to be run every minute or so, via cron.
 """
 import collections
+import datetime
 import json
 import http.client
 import logging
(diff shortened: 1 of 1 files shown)


MiguelCastillo (approved):
This seems like a reasonable approach

PR #326, opened 2025-05-27
[🔥AUDIT🔥] Fix directory references in e2e-test.groovy.

csilvers:
🖍 _This is an audit!_ 🖍

## Summary:
In Khan/webapp@#32285, some files got renamed that were called from
this script, but the script wasn't updated to match.  The e2e tests
have all been failing since that time.  This is an attempt to fix them
up.

I don't know why only some of the files called by this script were
moved to a new location, but not others.  Perhaps the rest will be
renamed at some future time.

Issue: https://khanacademy.slack.com/archives/C096UP7D0/p1748308879995339?thread_ts=1748289033.662619&cid=C096UP7D0

## Test plan:
I ran this new code via "replay" at
   https://jenkins.khanacademy.org/job/deploy/job/e2e-test/76145/console

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -190,6 +190,9 @@ def initializeGlobals() {
 def _setupWebapp() {
    kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_SHA1);
 
+   dir("webapp/testing/e2e") {
+      sh("make npm_deps");
+   }
    dir("webapp/services/static") {
       sh("make npm_deps");
    }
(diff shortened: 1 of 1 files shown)


nathanjd (approved):
Thanks for tracking this down!

jeresig (approved):
Ahh - so we're in the process of removing the test running logic here, and moving to running the tests on Github. Didn't realize that this specific bit was still being used in both the old and new setup, thank you for fixing this! cc: @michaelmendoza42 

I'll land this now to fix up the queue.

michaelmendoza42:
thanks for fixing this! sorry i forgot to update this script after i landed my changes.

PR #327, opened 2025-05-28
changed script to use github workflow

michaelmendoza42:
this is tied to the pr here:
https://github.com/Khan/webapp/pull/32317

the script starts a github workflow, waits for results, and sends a slack notification

ran the script on custom jenkins job:
https://jenkins.khanacademy.org/job/misc/job/e2e-gh-actions/78/console

### DIFF
--- a/jobs/e2e-test.groovy
+++ b/jobs/e2e-test.groovy
@@ -2,8 +2,10 @@
 //
 // cypress e2e tests are the smoketests run in the webapp/feature/cypress repo,
 // that hit a live website using lambdatest cli.
+
 @Library("kautils")
 // Classes we use, under jenkins-jobs/src/.
+import groovy.json.JsonBuilder;
 import org.khanacademy.Setup;
 // Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
 //import vars.withVirtualenv
@@ -135,10 +137,6 @@ for more information.""",
 
 ).apply();
 
-// We set these to real values first thing below; but we do it within
-// the notify() so if there's an error setting them we notify on slack.
-NUM_WORKER_MACHINES = null;
-
 // Override the build name by the info that is passed in (from buildmaster).
 REVISION_DESCRIPTION = params.REVISION_DESCRIPTION ?: params.GIT_REVISION;
 // Drop this part so the branches can be grouped in Cypress Cloud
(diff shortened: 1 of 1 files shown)


jobs/e2e-test.groovy
```diff
@@ -157,215 +153,82 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
...
-   }
-}
+GITHUB_TOKEN = null;
 
-def initializeGlobals() {
-   NUM_WORKER_MACHINES = params.NUM_WORKER_MACHINES.toInteger();
+SLACK_TOKEN = null;
+
+// Use first worker available
+WORKER_TYPE = "ka-firstinqueue-ec2";
```
nathanjd, line 161:
```suggestion
// We have a dedicated set of workers for the second smoke test.
WORKER_TYPE = (params.USE_FIRSTINQUEUE_WORKERS
               ? 'ka-firstinqueue-ec2' : 'ka-test-ec2');
```

This needs to only use the `ka-firstinqueue-ec2` when actually first in queue or else we'll slow down deploys. Here is where buildmaster sends the parameter.

https://github.com/Khan/buildmaster2/blob/0cbf0079b553352f765ebef387e9c44b2c379e32/src/state_machines.py#L909-L921

jobs/e2e-test.groovy
```diff
@@ -157,215 +153,82 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
...
-            }
+   ]);
+}
 
-            // Let notify() know not to send any messages to slack, because we
-            // just did it here.
-            env.SENT_TO_SLACK = '1';
-         }
+def withGitHubToken(body) {
+   wrap([$class: 'MaskPasswordsBuildWrapper', varPasswordPairs: [[password: GITHUB_TOKEN]]]) {
```
nathanjd, line 181: Nice!

nathanjd (changes requested):
Love the simplification! Just requesting one change to restore the old behavior for using the ka-firstinqueue-ec2 workers.

jobs/e2e-test.groovy
michaelmendoza42, line 161: not sure what this should actually be set to

jobs/e2e-test.groovy
```diff
@@ -157,215 +153,82 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
...
 
 onWorker(WORKER_TYPE, '5h') {     // timeout
    notify([slack: [channel: params.SLACK_CHANNEL,
-                   thread: params.SLACK_THREAD,
-                   sender: 'Testing Turtle',
-                   emoji: ':turtle:',
-                   when: ['FAILURE', 'UNSTABLE']],
-           buildmaster: [sha: params.GIT_REVISION,
-                         what: E2E_RUN_TYPE]]) {
+                  thread: params.SLACK_THREAD,
```
michaelmendoza42, line 198: i think this is the correct spacing now?

nathanjd: It's hard to say given it's not completely consistent across jobs, but our most used and maintained jobs, deploy-webapp, webapp-test, and merge-branches all align `channel` with `thread`, and `slack` with `buildmaster`, for example. So how it was before this change would be correct.

But I honestly don't care to enforce formatting that we don't lint or have a style guide for.

jobs/e2e-test.groovy
```diff
@@ -157,215 +153,86 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
...
 
+def _pullWebapp() {
+   // need script to start the workflow
+   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_SHA1, [], force=true);
+}
+
 // Determines if we are running the first or second smoke test.
-E2E_RUN_TYPE = (E2E_URL == "https://www.khanacademy.org" ? "second-smoke-test" : "first-smoke-test");
+IS_PRODUCTION = (E2E_URL == "https://www.khanacademy.org");
+E2E_RUN_TYPE = IS_PRODUCTION ? "second-smoke-test" : "first-smoke-test";
```
nathanjd, line 196: Thanks for catching this!

nathanjd (approved):
Our world just got a little less groovy and I love it.

jobs/e2e-test.groovy
csilvers: Did this require you to add the mask-passwords jenkins plugin?  If so, we should check in that requirement in code -- see Khan/aws-config:jenkins/README.plugins.md

Also, while this masks the password in the console, it doesn't mask it in the actual environment (of course), so someone with access to a jenkins worker machine could see the secret while tihs job is running.  That's not the world's worst security threat, but it's not awesome either.  I try really hard not to pass secrets via envvars. But perhaps it's unavoidable here?

jobs/e2e-test.groovy
```diff
@@ -157,215 +153,86 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
...
+   wrap([$class: 'MaskPasswordsBuildWrapper', varPasswordPairs: [[password: GITHUB_TOKEN]]]) {
+      withEnv(["GITHUB_TOKEN=${GITHUB_TOKEN}"]) {
+         body()
       }
    }
 }
 
+def _pullWebapp() {
+   // need script to start the workflow
+   kaGit.safeSyncToOrigin("git@github.com:Khan/webapp", GIT_SHA1, [], force=true);
```
csilvers, line 191: Why do you have `force=true` here?

jobs/e2e-test.groovy
```diff
@@ -157,215 +153,86 @@ DEPLOYER_USER = params.DEPLOYER_USERNAME.replace("@", "")
...
-      }
+      _pullWebapp();
       stage("Run e2e tests") {
-         withEnv([
-            "COMMIT_INFO_BRANCH=${SHORT_REVISION_DESCRIPTION}",
-            "KA_SKIP_GEN=1",
-         ]) {
-            runAllTestClients();
+         withGitHubToken {
+            withEnv(["SLACK_TOKEN=${SLACK_TOKEN}"]) {
```
csilvers, line 211: In this case, you could avoid having to read the secret via an envvar by having notify-workflow-status.ts load the secret itself.  (Either using the js gsm library, or just calling out to gcloud in a subprocess.) 
 Then that script won't depend on envvars as much either, which is a maintainability win.

PR #332, opened 2025-05-30
[🔥AUDIT🔥] fix checkout ref to use git_sha

michaelmendoza42:
🖍 _This is an audit!_ 🖍

## Summary:
fix checkout ref to use git_sha

Issue: INFRA-XXXX

## Test plan:


csilvers (commented):
Definitely better!

PR #334, opened 2025-05-30
[🔥AUDIT🔥] Fix the REPORT_DIR location.

jeresig:
🖍 _This is an audit!_ 🖍

## Summary:
This got changed in webapp and Jenkins is getting tripped up here.

Issue: FEI-XXXX

## Test plan:


csilvers (commented):
I don't really understand this change, I'm afraid.  services/static/genfiles is intended for use by code that lives in services/static; it will go away when services/static does.  But this jenkins job isn't going away when services/static goes away, right?  So won't this start to break at that point?

jeresig:
@csilvers Yep, that's correct - this PR was a transitional fix to get the e2e tests working again. It was reverted immediately after when we transitioned to using `testing/e2e` completely (and is no longer in webapp).

PR #343, opened 2025-07-08
Use `input` instead of a deploy param, to control deploying to production.

csilvers:
## Summary:
build-and-deploy-publish-worker.groovy has an input parameter to say
whether it does just-build, or build-and-deploy.  This PR changes it
to use an interactive prompt instead, just like we do with
deploy-fastly.groovy.  That way, a user can look at the output of the
build step and decide if it looks good or not, and then say whether to
deploy or not based on that.

Issue: https://github.com/Khan/jenkins-jobs/pull/323#discussion_r2089652443

## Test plan:
Fingers crossed

### DIFF
--- a/jobs/build-and-deploy-publish-worker.groovy
+++ b/jobs/build-and-deploy-publish-worker.groovy
@@ -33,11 +33,6 @@ WITH CAUTION!</b> -- only for situations where you are overwriting
 someone else's publish-deploy on purpose.""",
     true
 
-).addBooleanParam(
-    "DEPLOY_TO_PROD",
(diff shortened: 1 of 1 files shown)


csilvers:
@zhiyi-zhang-duke I had mentioned this when reviewing your script originally about a month ago.  This is what it would look like -- feel free to try it out!  I don't know if it matches your team's workflow well or not.

MiguelCastillo (approved):
Good idea.

zhiyi-zhang-duke:
Yes that makes sense to me, I'll be sure to update the docs. Please ping me when this is merged so I can update the jenkins job!

zhiyi-zhang-duke (approved):

PR #347, opened 2025-07-26
Make buildmaster update message more clear

nathanjd:
## Summary:
Some folks have been getting confused by the following message in the
merge-branches job logs:

```
17:22:43  Marking commit #12345 as failed: null
```

This leads users to think this is an error message with a null reason as
opposed to what it actually is: a log of the status update we're sending
to buildmaster.

This happens whenever merge-branches encounters a failure, such as a
merge conflict or, in the reported case, when an commit-ish was passed
that doesn't exist in origin. The actual error message exists at the
bottom of the console logs and in the pipeline steps view next to the
red (X).

Change this message to:

```
17:53:27  Marking buildmaster commit #12345 as failed: No merged SHA
```

Issue: https://khanacademy.atlassian.net/browse/INFRA-10684

Test plan:

Tested with the replay feature:

https://jenkins.khanacademy.org/job/deploy/job/merge-branches/324339/console

### DIFF
--- a/vars/buildmaster.groovy
+++ b/vars/buildmaster.groovy
@@ -124,7 +124,8 @@ def notifyStatus(job, result, sha1) {
 }
 
 def notifyMergeResult(commitId, result, sha1, gae_version_name) {
-   echo("Marking commit #${commitId} as ${result}: ${sha1}");
+   String shaToLog = sha1 ? sha1 : "No merged SHA"
+   echo("Marking buildmaster commit #${commitId} as ${result}: ${shaToLog}");
    def params = [
       commit_id: commitId,
       result: result,


MiguelCastillo (commented):
Ohh!  I was thinking of something a lil bit differently, which I realize what Im asking for in the ticket is very unclear.  I didnt even know exactly how I wanted this addressed, but I think i understand that better now.

As you point out in the ticket, we already log `Cannot find 'deploy/non-existent' in repo 'git@github.com:Khan/webapp'` here https://github.com/Khan/jenkins-jobs/blob/0670bc5a1c0bab364dfb981f7854b6b17bdd49db/vars/kaGit.groovy#L87.  That's exactly the message I wanted.  But in my mind the issue is that it makes it into jenkins logs and in the jenkins UI.  Both places aren't particularly intuitive for engineers to use for debugging.  Often what we show in the Jenkins workflow UI can annoying to track because we often have multiple ambiguous red checkmarks.

What I'd love to see is actually the reason why we had a failure right in slack.  See screenshot below for the example of me queueing a branch that doesn't exist.  When Kevin ran into this, I had to go to jenkins to figure out why this failed.

My hope is that we can provide our engineers with enough details right in slack so that engineers don't have to go to jenkins.  While this might not be possible to do 100% of cases, I'd love to have as a goal is reduce the trips our engineers need to make to jenkins.

FWIW - BenK did have a TODO example for what Im asking for here https://github.com/Khan/jenkins-jobs/blob/0670bc5a1c0bab364dfb981f7854b6b17bdd49db/vars/kaGit.groovy#L244.

Screenshot 2025-07-30 at 10 40 30 AM

https://khanacademy.slack.com/archives/C096UP7D0/p1753886275263109

I would love for us to consider no longer giving folks jenkins links and telling them "(look for the red circle(s))" as much as possible as a way to debug deploy issues, whenever possible.

MiguelCastillo (approved):
Ok I see what we are trying to do in this PR.  I left a comment inline that explain a bit more the exact pain point Im referring in the ticket in the first place.  The addition of providing more details in slack is in addition to improving logs in jenkins.  Happy to tackle them all in separate PRs.

csilvers:
It sounds like Miguel has a particular vision for this PR, so I'll leave reviewing to him!

nathanjd:
Merge failure output is sent to slack with this PR:

https://github.com/Khan/jenkins-jobs/pull/349

PR #373, opened 2026-02-10
Build all our dockerfile images in parallel.

csilvers:
## Summary:
Now that the builds are done in the cloud rather than on the local
machine, we can do them in parallel, saving a bunch of time!  And we
can run the job on master instead of needing to spin up a worker,
since now it hardly takes up any (local) resources at all.

Issue: none

## Test plan:
I ran the new script with success, by hitting "replay" on an
existing job and pasting in the new code into the "main"
section of the replay page.  Results:
   https://jenkins.khanacademy.org/job/misc/job/update-devserver-static-images/1032/

### DIFF
--- a/jobs/update-devserver-static-images.groovy
+++ b/jobs/update-devserver-static-images.groovy
@@ -64,13 +64,17 @@ def _setupWebapp() {
    }
 }
 
-def uploadService(service) {
+def uploadServices(servicesString) {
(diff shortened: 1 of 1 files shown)


nathanjd (approved):
Love it! I'm excited to see if this ends up making build-webapp faster.

PR #374, opened 2026-03-04
Update the build-current-sqlite job to use Firestore

dnerdy:
## Summary:
This PR updates the build-current-sqlite job to use the Firestore emulator. The
dev server has been updated to support starting with the Firestore emulator in
https://github.com/Khan/webapp/pull/38040.

Issue: INFRA-10405

## Test plan:
I ran the following command on my local machine:
```
docker exec webapp-datastore-emulator-1 dash -c 'ls /proc/*/cmdline | while read cmd; do echo -n "$cmd " | cat - "$cmd" 2>/dev/null | tr "\0" " " | awk "{ print }" | grep "/usr/bin/qemu-x86_64 /usr/bin/firestore-emulator-plus" | grep -v dash | cut -d/ -f3 | while read pid; do kill $pid; done; done'
```
and verified it caused the firestore-emulator-plus wrapper to export the
emulator data to datastore/dev_datastore.tar.gz by inspecting the
datastore-emulator container logs in Docker Desktop.

Otherwise, 🤞.

### DIFF
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -57,26 +57,9 @@ def runScript() {
 
          updateDatabase();
 
-         // Send SIGTERM to the firestore emulator to have it write out its
-         // backing file.
-         //
-         // TODO(marksandstrom) Is there a better, more direct way to do this?
-         // We want to send SIGTERM to the process running the command
-         // /usr/bin/firestore-emulator-plus.
-         //
-         // The command works as follows:
-         // - list the command lines of the running processes
-         //       ls /proc/*/cmdline
-         // - echo the /proc/ entry and the command line on the same line
-         //       echo -n "$cmd " | cat - $cmd
-         // - replace the \0 characters (from cmdline) with spaces
-         // - add a newline to the end of the cmdline output
-         //       awk '{ print }'
-         // - find the /usr/bin/firestore-emulator-plus process using grep
-         // - extract the pid from the /proc/ entry -- cut -d/ -f3
-         // - then, if there's output -- while read pid
-         // - send SIGTERM to the pid -- kill $pid
-         sh("docker exec webapp-datastore-emulator-1 dash -c 'ls /proc/*/cmdline | while read cmd; do echo -n \"$cmd \" | cat - \"$cmd\" 2>/dev/null | tr \"\\0\" \" \" | awk \"{ print }\" | grep \"/usr/bin/qemu-x86_64 /usr/bin/firestore-emulator-plus\" | grep -v dash | cut -d/ -f3 | while read pid; do kill $pid; done; done'")
+         // Send SIGTERM to the firestore-emulator-plus wrapper to have
+         // Firestore write out its backing file. The wrapper runs as PID 1.
+         sh("docker exec webapp-datastore-emulator-1 dash -c 'kill 1'")
 
          // Now upload the new database to gcs.
          sh("gsutil cp ${params.CURRENT_SQLITE_BUCKET}/dev_datastore.tar.gz ${params.CURRENT_SQLITE_BUCKET}/dev_datastore.tar.gz.bak");
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -53,7 +53,7 @@ def runScript() {
          // need the `env` because otherwise jenkins's BUILD_TAG
          // takes precedence over our own.
          // TODO(csilvers): clear more of env?
-         sh("ssh-agent env -u BUILD_TAG USE_FIRESTORE_EMULATOR=true make start-dev-server-backend WORKING_ON=NONE");
+         sh("ssh-agent env -u BUILD_TAG make start-dev-server-backend WORKING_ON=NONE");
 
          updateDatabase();
 
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -53,7 +53,7 @@ def runScript() {
          // need the `env` because otherwise jenkins's BUILD_TAG
          // takes precedence over our own.
          // TODO(csilvers): clear more of env?
-         sh("ssh-agent env -u BUILD_TAG USE_FIRESTORE_EMULATOR=true SKIP_FIRESTORE_SNAPSHOT_DOWNLOAD=true make start-dev-server-backend WORKING_ON=NONE");
+         sh("ssh-agent env -u BUILD_TAG USE_FIRESTORE_EMULATOR=true make start-dev-server-backend WORKING_ON=NONE");
 
          updateDatabase();
 
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -44,9 +44,8 @@ def runScript() {
    kaGit.safeSyncToOrigin("git@github.com:Khan/frontend", "main");
 
    dir("webapp") {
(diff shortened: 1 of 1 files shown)


jobs/build-current-sqlite.groovy
```diff
@@ -44,29 +44,43 @@ def runScript() {
...
 
          updateDatabase();
 
-         // I don't know a better way to flush the datastore
-         // changes to disk, than to just kill it.
-         sh("docker exec webapp-datastore-emulator-1 dash -c 'kill `ls -l /proc/*/exe | grep java- | cut -d/ -f3`'")
+         // Send SIGTERM to the firestore emulator to have it write out its
+         // backing file.
+         //
+         // TODO(marksandstrom) Is there a better, more direct way to do this?
```
csilvers: I thought the wrapper you wrote supported writing out the datastore when you send it `/quit`?

jobs/build-current-sqlite.groovy
```diff
@@ -44,29 +44,43 @@ def runScript() {
...
+         // - echo the /proc/ entry and the command line on the same line
+         //       echo -n "$cmd " | cat - $cmd
+         // - replace the \0 characters (from cmdline) with spaces
+         // - add a newline to the end of the cmdline output
+         //       awk '{ print }'
+         // - find the /usr/bin/firestore-emulator-plus process using grep
+         // - extract the pid from the /proc/ entry -- cut -d/ -f3
+         // - then, if there's output -- while read pid
+         // - send SIGTERM to the pid -- kill $pid
+         sh("docker exec webapp-datastore-emulator-1 dash -c 'ls /proc/*/cmdline | while read cmd; do echo -n \"$cmd \" | cat - \"$cmd\" 2>/dev/null | tr \"\\0\" \" \" | awk \"{ print }\" | grep \"/usr/bin/qemu-x86_64 /usr/bin/firestore-emulator-plus\" | grep -v dash | cut -d/ -f3 | while read pid; do kill $pid; done; done'")
```
csilvers: Out of curiousity: Why is it using qemu?  We shouldn't need to emulate when this is running on jenkins, no?

csilvers (commented):
I think this is ok?, though I'm not sure why the old kill command doesn't work anymore.  Also, I'm not sure about qemu.

But I am wondering if we want the USE_FIRESTORE_EMULATOR envvar at all.  Since you made the original upload of it already.  Now we can just wait to land the change that switches the devserver to use firestore emulator, and then check in a version here that doesn't need to set the envvar.  I think maybe it will be simpler? but I'm not sure.

jobs/build-current-sqlite.groovy
dnerdy: The Firestore emulator persists data when sending a POST to the /shutdown endpoint (but the emulator hands after writing out the data and doesn't shutdown cleanly). The wrapper calls /shutdown when it receives SIGTERM and then sends a SIGTERM to the emulator once it sees that it's written out the data (and hangs).

jobs/build-current-sqlite.groovy
dnerdy: `ls -l /proc/*/exe` was just listing pids for me, e.g. `/proc/365/exe`, so there's no information available to pick out the correct process. When looking at the content of `/proc/*/cmdline` all process commands where `/usr/bin/qemu-x86_64` (with `/usr/bin/firestore-emulator-plus` passed as the first argument for the process we're interested in sending SIGTERM to). Maybe this has to do with running the image on macOS, though? I assumed the processes would look the same in the container regardless of where the container is run, but maybe not.

dnerdy: @csilvers what do you think about the idea of updating the wrapper to support a `--write-pid /tmp/firestore-emulator-plus-pid` option? Then sending a SIGTERM to the process would be super simple.

csilvers: > ls -l /proc/*/exe was just listing pids for me

Interesting, for me (on linux) they are symlinks:
```
lrwxrwxrwx 1 csilvers             csilvers             0 Mar  1 17:09 /proc/8329/exe -> /usr/bin/xeyes*
```
Not all of them are symlinks though, I'm not exactly sure which are.  Did you check the full output of the `ls -l` to make sure it was just pid's for everything?

csilvers: > @csilvers what do you think about the idea of updating the wrapper to support a --write-pid /tmp/firestore-emulator-plus-pid option? Then sending a SIGTERM to the process would be super simple.

I'm not opposed to that, but it seems like it shouldn't be necessary.  Remind me why we can't just use `ps` for this?  Is it not installed on the image?  Maybe a better solution is to just install busybox in the image.

dnerdy: That's right. The image doesn't contain `ps`.

dnerdy: > > ls -l /proc/*/exe was just listing pids for me
> 
> Interesting, for me (on linux) they are symlinks:
> 
> ```
> lrwxrwxrwx 1 csilvers             csilvers             0 Mar  1 17:09 /proc/8329/exe -> /usr/bin/xeyes*
> ```
> 
> Not all of them are symlinks though, I'm not exactly sure which are. Did you check the full output of the `ls -l` to make sure it was just pid's for everything?

Ah, I was misremembering slightly. It does include the command, but as mentioned above that isn't useful:

```
# ls -l /proc/*/exe
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/1/exe -> /usr/bin/qemu-x86_64
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/15/exe -> /usr/bin/qemu-x86_64
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/90/exe -> /usr/bin/qemu-x86_64
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/96/exe -> /usr/bin/qemu-x86_64
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/986/exe -> /usr/bin/qemu-x86_64
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/self/exe -> /usr/bin/ls
lrwxrwxrwx 1 root root 0 Mar  5 17:58 /proc/thread-self/exe -> /usr/bin/qemu-x86_64
```

csilvers: Ah, that _is_ a mac vs linux thing -- because you're on (arm) mac everything has to run through the emulator.

So I think any of these approach would be fine:
1. Install busybox in the image so we can use ps
2. Do the complicated proc processing like you're doing now
3. Print the pid to a pidfile in the binary

You can pick which is your favorite!

csilvers: Or you could ignore it and leave the current code, since jenkins runs on linux so the existing `ps` code in the jenkins job will work correctly there.

dnerdy: Actually, with the exec change you made in https://github.com/Khan/webapp/pull/38051 the wrapper will now always be pid 1. So, we can now just change this to `kill 1`.

csilvers: Heh!

csilvers: I don't think that's true, but we can check.  The `exec` is replacing the `sh`, not the `init`.

dnerdy: The sh is the init, though (since it's the entrypoint).

csilvers (approved):
Beautiful!

PR #375, opened 2026-03-17
Add lockable resource to prevent update-traffic timeout

nathanjd:
## Summary:
We have been running into a particularly frusting issue a few times a
day for the last month: the update-traffic command will hang until its
60 minute timeout despite the traffic migration being complete 5-10
minutes in. We've received a preliminary RCA from Google's engineering
teams:

1. When the update-traffic command is kicked off for a given service, it
   creates an intent that expects the traffic allocation across all
   revisions (with traffic tags or higher than 0% traffic allocation) to
   achieve a specific target state.
2. A DeleteRevision is kicked off for the same service that deletes one
   of the revisions specified in the target traffic allocation state.
3. As a result, the service-level intent for the update-traffic command
   can never be fulfilled. The command then polls until timeout.
   
https://khanacademy.slack.com/archives/C027VGTEVAA/p1773670531790489?thread_ts=1772132199.620719&cid=C027VGTEVAA

Prevent this from happening by having both set-default and
delete-versions share a lockable resource, `update-traffic-lock`. Set
priority higher for set-default as we want to minimize the slowdown the
lock may introduce to our deploys.

The lockable resource has been created in jenkins at:

https://jenkins.khanacademy.org/manage/configure

Screenshot 2026-03-16 at 5 05 14 PM

Issue: https://khanacademy.atlassian.net/browse/INFRA-10970

Test plan:

1. Monitor the deployment queue.
2. Every day the issue does not happen, report it to this slack thread:
   https://khanacademy.slack.com/archives/C027VGTEVAA/p1772132199620719
3. If the issue does happen again, report to the same thread for
   advisement.
4. Once no occurances of the issue have been reported for a week, close
   the thread with Google.

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -954,7 +954,21 @@ onMaster('4h') {
          try {
             stage("Promoting and monitoring") {
                withVirtualenv.python3() {
-                  setDefaultAndMonitor();
+                  // Wait for a concurrent delete-version delete stage to finish
+                  // before performing any traffic migrations. Use high priority
+                  // to jump in front of any lock-queued but not-yet-in-progress
+                  // delete stages.
+                  //
+                  // We do this because the update-traffic command creates a
+                  // LongRunningOperation that expects a specific target traffic
+                  // allocation across all revisions with traffic tags and/or a
+                  // traffic % higher than 0%. If a revision is deleted while
+                  // the traffic migration is in-progress, the target state will
+                  // never be achieved so the operation will wait until its full
+                  // 60 minute timeout.
+                  lock(resource: 'update-traffic-lock', priority: 20) {
+                     setDefaultAndMonitor();
+                  }
                }
             }
             // Unlike above, we do not need to verify second smoke tests
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -87,9 +87,31 @@ onMaster('30m') {
       stage("Initializing webapp") {
          _setupWebapp();
       }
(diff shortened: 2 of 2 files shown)


MiguelCastillo (approved):
This seems very reasonable.  let's give it a try.

jobs/deploy-webapp.groovy
```diff
@@ -954,7 +954,21 @@ onMaster('4h') {
...
+                  // delete stages.
+                  //
+                  // We do this because the update-traffic command creates a
+                  // LongRunningOperation that expects a specific target traffic
+                  // allocation across all revisions with traffic tags and/or a
+                  // traffic % higher than 0%. If a revision is deleted while
+                  // the traffic migration is in-progress, the target state will
+                  // never be achieved so the operation will wait until its full
+                  // 60 minute timeout.
+                  lock(resource: 'update-traffic-lock', priority: 20) {
```
MiguelCastillo, line 969: Im not sure it makes a difference.  But in the delete job, the lock wraps the `stage`.  In this deploy-webapp, it does not.

nathanjd: Good point! I'll make it consistent. I think I'd rather have it report that it's within a given stage while waiting for the lock.

nathanjd: Updated.

csilvers (approved):
Great thinking to set the priority.

Once this is done, please update Khan/aws-config:jenkins/README.plugins.md (it will need to be in a separate PR, obviously), to list the plugin.

PR #376, opened 2026-04-01
Properly handle branches with slashes.

csilvers:
## Summary:
I can't believe this hasn't gotten us until trouble before!  I guess
because a) we don't use submodules anymore, and b) most of our jobs
don't try to push to a branch except for `master`, which doesn't have
a slash in it.  But some jobs do, and we ran into trouble there.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1775057192305139

## Test plan:
I ran:
```
git rev-parse --symbolic-full-name HEAD | sed -e 's,^refs/heads/,,' -e 's,^refs/remotes/,,' -e 's,^refs/tags/,,'
```
locally and it emitted `csilvers`, same as before this change.

### DIFF
--- a/safe_git.sh
+++ b/safe_git.sh
@@ -318,7 +318,7 @@ pull() {
 push() {
     (
     cd "$1"
-    branch=`git rev-parse --symbolic-full-name HEAD | sed 's,^.*/,,'`
+    branch=`git rev-parse --symbolic-full-name HEAD | sed -e 's,^refs/heads/,,' -e 's,^refs/remotes/,,' -e 's,^refs/tags/,,'`
     # In case there have been any changes since the script began, we
     # do 'pull; push'.  On failure, we undo all our work.
     _fetch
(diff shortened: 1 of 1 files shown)


safe_git.sh
```diff
@@ -318,7 +318,7 @@ pull() {
 push() {
     (
     cd "$1"
-    branch=`git rev-parse --symbolic-full-name HEAD | sed 's,^.*/,,'`
+    branch=`git rev-parse --symbolic-full-name HEAD | sed -e 's,^refs/heads/,,' -e 's,^refs/remotes/,,' -e 's,^refs/tags/,,'`
```
kevinb-khan, line 321: `git branch --show-current` might be a better approach since it avoids needing to strip `refs/{remotes,heads,tags,...}/`.

kevinb-khan (approved):
LGTM.  Thanks for updating this to handle branch names with `/` in them.

safe_git.sh
csilvers: You're probably right!  Most of our code needs to work on any commit-ish, but if you're pushing it probably needs to be a branch.  It's a bit scarier change though.  I'll leave a TODO.

csilvers: Though I don't know what that does in subrepos.  Maybe I'll just keep it as-is.

MiguelCastillo:
This fix works for me.  Thank you!

On the upshot, I will be very glad to work towards getting rid of scripts as we start moving towards merge queues.  So Im inclined to keep changes to as minimal as possible here.

PR #378, opened 2026-04-15
Let the docker command succeed in non-interactive-land.

csilvers:
## Summary:
The build-current-sqlite jenkins job has been failing for the last
several weeks.  Looking into it, the problem is in running
`docker exec -it`:
```
cannot attach stdin to a TTY-enabled container because stdin is not a terminal
```

I just remove the `-it`, along the lines of the other `docker exec`
call in this file.  Hopefully that is enough to get things working
again.

Issue: https://jenkins.khanacademy.org/job/misc/job/build-current-sqlite/1610/console

## Test plan:
Fingers crossed

### DIFF
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -59,11 +59,14 @@ def runScript() {
 
          // Send SIGTERM to the firestore-emulator-plus wrapper to have
          // Firestore write out its backing file. The wrapper runs as PID 1.
+         // NOTE: we can't use `docker exec -it` since we are not attached
+         // a terminal and are not interactive.  But we can run dash instead
+         // and get the same effect.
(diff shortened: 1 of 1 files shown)


MiguelCastillo (approved):
Good catch!

jobs/build-current-sqlite.groovy
```diff
@@ -59,11 +59,14 @@ def runScript() {
 
          // Send SIGTERM to the firestore-emulator-plus wrapper to have
          // Firestore write out its backing file. The wrapper runs as PID 1.
+         // NOTE: we can't use `docker exec -it` since we are not attached
+         // a terminal and are not interactive.  But we can run dash instead
+         // and get the same effect.
```
MiguelCastillo, line 64: TIL about dash

PR #381, opened 2026-04-20
Use the new syntax for should_deploy.py and trivial_diffs.py.

csilvers:
## Summary:
These scripts used to take a from-commit and to-commit, and would
internally combine them to "<from_commit>...<to_commit>".  I changed
the script to accept the actual commit-range explicitly instead.  Now
I'm changing jenkins to use that new format.

I didn't change jenkins earlier because I didn't want to break znd's
that didn't have the updates to should_deploy.py and trivial_diffs.py.
But those files were updated a month ago, and I don't expect any znd's
to be based on such old code, so I think it's safe to do this now!

Issue: none

## Test plan:
Fingers crossed


MiguelCastillo (approved):
Great!

PR #382, opened 2026-04-28
Add a test for the merge-branches job

jaredly:
## Summary:
This is to prepare for migrating the job to a webapp github action. I wanted there to be a test that could run both on the groovy and on the new go job, to verify that theyb behave the same.

codex did most of it.

Issue: none

## Test plan:
`./tests/merge_branches_integration_test.sh`


jaredly:
going a different direction

PR #384, opened 2026-05-01
Bridge the webapp-test job to github actions

jaredly:
## Summary:
Going with an incremental approach. Default is no change, but if you set USE_GITHUB_BRIDGE it will delegate to a github actions workflow.

Issue: https://khanacademy.atlassian.net/browse/INFRA-11003

## Test plan:
🤞 I think we'll have to deploy this to really test it out


vars/ghActions.groovy
```diff
@@ -0,0 +1,67 @@
...
+//   ref      - Branch name to run the workflow on
+//   headSha  - Resolved SHA for the ref (used to identify the new run)
+//   inputs   - Map of string→string workflow inputs (optional)
+//
+// Dispatches the workflow via the GitHub API, then polls for up to 30s
+// to locate the new run and return its ID.
+def dispatchGithubActionsWorkflow(Map args) {
+    def payload = JsonOutput.toJson([ref: args.ref, inputs: args.inputs ?: [:]])
+
+    withCredentials([string(credentialsId: 'GITHUB_TOKEN', variable: 'GH_TOKEN')]) {
```
csilvers: We prefer not to use jenkins credentials, since historically security there has not been great.  Can you see what it's like to add this to vars/withSecrets.groovy instead?  It may require giving the jenkins deploy script access to the github token gsm secret.

vars/ghActions.groovy
```diff
@@ -0,0 +1,67 @@
...
+        sh(script: """
```
csilvers: Use exec instead of sh, it will handle the shell-escaping for you, which this doesn't do as-is.

vars/ghActions.groovy
```diff
@@ -0,0 +1,67 @@
...
+        """)
+
+        // Poll for up to 30s (10 attempts × 3s) to find the new run.
+        // Filter by head_sha when available to avoid picking up a concurrent
+        // dispatch against the same branch.
+        def shaFilter = args.headSha ? "&head_sha=${args.headSha}" : ""
+        def runId = null
+        for (def i = 0; i < 10; i++) {
+            sleep(3)
+            def runsOutput = sh(script: """
```
csilvers: Here you can use exec.outputOf().

vars/ghActions.groovy
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+// Wait for a GitHub Actions workflow run to complete.
+// Blocks until the run finishes; fails the build if the run fails.
+//
+// runId - the workflow run ID (string or long)
+def waitForGithubActionsWorkflow(String runId) {
+    withCredentials([string(credentialsId: 'GITHUB_TOKEN', variable: 'GH_TOKEN')]) {
+        sh("gh run watch ${runId} --exit-status")
```
csilvers: (Use exec here too.)

vars/ghActions.groovy
csilvers, line 1: It seems like it will be pretty common to do `dispatchGithubActionsWorkflow` followed immediately by `waitForGithubActionsWorkflow`.  Maybe do something like:
```
def call(Map args) {
   def runID = dispatchGithubActionsWorkflow(args)
   waitForGithubActionsWorkflow(runID)
}
```
and then webapp-test can just call `ghActions(...)`.

csilvers: And maybe it would be more descriptive to rename `ghActions.groovy` to `runGithubAction.groovy`.

csilvers (changes requested):
I like the general idea!  Using a flag for this I think is key since I remember we had decided to use jenkins for deploy-tests for a while due to trust issues.  @jeresig may remember the details here better, and say whether they still apply.

Request Changes for my comments below, most notably making sure to do shell-escaping.  But it would also be good if you could add a ghActions.txt file that documents the new vars!

csilvers:
As part of this change, can you make the webapp-test run from jobs/make-allcheck.groovy set the new variable to true, so we run the nightly tests on github actions?  That should address https://khanacademy.slack.com/archives/C090HCPAFEU/p1777400545082369.

vars/withSecrets.groovy
```diff
@@ -11,6 +11,16 @@ import groovy.transform.Field
 @Field _activeSlackAndStackdriverSecretsBlocks = 0;
 @Field _activeGithubSecretsBlocks = 0;
 
+// Fetches the khan-actions-bot personal access token from GSM and
+// exposes it as the GITHUB_TOKEN env var for the duration of the closure.
+// Used by runGithubAction to authenticate GitHub API calls and the gh CLI.
+def githubActionsDispatch(Closure body) {
+   def token = sh(script: "gcloud --project khan-academy secrets versions access latest --secret khan_actions_bot_github_personal_access_token__Github_Action_Workflow__repo__workflow__read_org__read_discussion__", returnStdout: true).trim();
+   withEnv(["GITHUB_TOKEN=${token}"]) {
```
csilvers: This isn't awesome because the env is publicly accessible to all other processes running on the jenkins machine.  Of course, having a file, like the other functions in this file, isn't great either.  But maybe at least add a TODO to make this more secure.

vars/withSecrets.groovy
```diff
@@ -11,6 +11,16 @@ import groovy.transform.Field
 @Field _activeSlackAndStackdriverSecretsBlocks = 0;
 @Field _activeGithubSecretsBlocks = 0;
 
+// Fetches the khan-actions-bot personal access token from GSM and
+// exposes it as the GITHUB_TOKEN env var for the duration of the closure.
+// Used by runGithubAction to authenticate GitHub API calls and the gh CLI.
+def githubActionsDispatch(Closure body) {
+   def token = sh(script: "gcloud --project khan-academy secrets versions access latest --secret khan_actions_bot_github_personal_access_token__Github_Action_Workflow__repo__workflow__read_org__read_discussion__", returnStdout: true).trim();
```
csilvers: You don't need to use `sh` here, you can use `exec.outputOf`.

csilvers (approved):
OK, let's do it!

vars/withSecrets.groovy
jaredly: hmmm oh yeah I can switch to just passing a variable around

PR #385, opened 2026-05-07
Fix ref calculation for github_bridge w/ webapp-test job

jaredly:
## Summary:
I was trying to test out the new github bridge, passing in all the same parameters buildmaster does but w/ the USE_GITHUB_BRIDGE flag on, and it didn't work, because buildmaster passes a SHA for GIT_REVISION. Which I should have anticipated. Fortunately, buildmaster passes a branch ref as REVISION_DESCRIPTION, so I can use that.

Issue: none

## Test plan:
🤞

### DIFF
--- a/jobs/webapp-test.groovy
+++ b/jobs/webapp-test.groovy
@@ -552,11 +552,12 @@ onWorker('ka-test-ec2', '5h') {     // timeout
                          what: 'webapp-test']]) {
       initializeGlobals();
 
-      if (params.USE_GITHUB_BRIDGE && !_isSha(params.GIT_REVISION)) {
+      if (params.USE_GITHUB_BRIDGE && (!_isSha(params.GIT_REVISION) || params.REVISION_DESCRIPTION)) {
+         def ref = _isSha(params.GIT_REVISION) ? params.REVISION_DESCRIPTION : params.GIT_REVISION
          runGithubAction(
             repo: "Khan/webapp",
             workflow: "webapp-test.yml",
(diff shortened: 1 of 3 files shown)


jobs/webapp-test.groovy
```diff
@@ -552,11 +552,12 @@ onWorker('ka-test-ec2', '5h') {     // timeout
                          what: 'webapp-test']]) {
       initializeGlobals();
 
-      if (params.USE_GITHUB_BRIDGE && !_isSha(params.GIT_REVISION)) {
+      if (params.USE_GITHUB_BRIDGE && (!_isSha(params.GIT_REVISION) || params.REVISION_DESCRIPTION)) {
+         def ref = _isSha(params.GIT_REVISION) ? params.REVISION_DESCRIPTION : params.GIT_REVISION
```
csilvers: I don't think this works, what if you have `sun: test branch1 + branch2`?

csilvers (changes requested):
> because buildmaster passes a SHA for GIT_REVISION

Maybe we can fix this in buildmaster?  I think it always creates a tag, maybe we can pass in the tag (assuming that works for github actions).

jaredly:
ah yeah that sounds better

csilvers (approved):
Nice -- definitely happy to be explicit about passing in the tag rather than using the `_isSha` check, too.

nathanjd (approved):
Looks good along with the changes from https://github.com/Khan/buildmaster2/pull/361/changes

PR #387, opened 2026-05-12
[🔥AUDIT🔥] Fix onMaster invocation

jaredly:
🖍 _This is an audit!_ 🖍

## Summary:
crazy that in 2026 we have to interact with a language that doesn't have static validation of method invocations 😭

turns out you can't just `onMaster {` you need to give it a timeout.

Issue: https://jenkins.khanacademy.org/job/deploy/job/webapp-test/340742/console

## Test plan:
🤞😭


csilvers (commented):
👍

PR #389, opened 2026-05-12
Don't set set-default-status to "started" after "finished"

nathanjd:
## Summary:
We've been observing that many attended deploys never receive the
finish-up prompt in slack despite the deploy-webapp job entering the
polling state for the finish-up prompt to be accepted. I've traced this
to set-default-status being set to "started" AFTER it is set to
"finished". As a result, the prompt is never sent in slack as it
requires the set-default-status to be "finished". However, it is set to
"finished" for long enough to kick off second smoke tests and proceed
with the rest of the deploy.

Occurrences of "started" after "finished".

- https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27898/consoleFull (deployed many services but had a hanging update-traffic command to the `assign-content` service)
- https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27892/consoleFull (this one was just `fastly-khanacademy-compute` which could explain why @MiguelCastillo has been hitting this so consistently.)
- https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27867/consoleFull (also just `fastly-khanacademy-compute`)

To fix this, remove all logic to "wait for set default start". With the
change to only start second smoke tests AFTER set default is completed,
we no longer have any process that needs to trigger at that moment.

https://github.com/Khan/buildmaster2/pull/359

Instead, just set it to "started" right before attempting the traffic
migration. This is a bit of a change in semantics to
`set-default-state == "started"`, but it is more accurate for the current
state of buildmaster. The monitoring script itself knows how to wait for
the traffic migration to actually begin.

This likely means we can remove deploy/wait_for_default.py from webapp
entirely after this change.

https://github.com/Khan/webapp/blob/9901303cdfe2f1d03051b132ab6a47ae0bd14bfd/deploy/wait_for_default.py

Issue: https://khanacademy.atlassian.net/browse/INFRA-11020

## Test plan:
1. Monitor production deploys in #1s-and-0s-deploys. https://khanacademy.slack.com/archives/C096UP7D0

### DIFF
--- a/jobs/deploy-webapp.groovy
+++ b/jobs/deploy-webapp.groovy
@@ -2,33 +2,38 @@
 
 // Sending traffic to a version is a complex process, and this is a complex
 // script.  For more high-level information on deploys, see
-//    https://docs.google.com/document/d/1Zr0wwzbvPkmN_BFAsrMPZhccIb0HHJUgmLZfBUWYFvA/edit
+// https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/360939604/User+s+Guide+to+the+Khan+Deployment+System
 // For more details on how this is used as a part of our build process, see the
-// buildmaster (github.com/Khan/buildmaster) and its design docs:
-//     https://docs.google.com/document/d/1utyUMMBOQvt4o3W_yl_89KdlNdAZUYsnSN_2FjWL-wA/edit#heading=h.kzjq9eunc7bh
+// buildmaster (https://github.com/Khan/buildmaster2) and its design docs:
+// https://docs.google.com/document/d/1utyUMMBOQvt4o3W_yl_89KdlNdAZUYsnSN_2FjWL-wA/edit#heading=h.kzjq9eunc7bh
 
-// By the time we run, a new version has already been uploaded to App Engine
-// and/or Google Cloud Storage with the relevant set of changes, end-to-end
-// tests have been run on it, and unit tests have been run on the corresponding
-// code.  Here's what we do, some of it in parallel:
+// By the time we run, new revisions(s) have already been uploaded to Cloud Run
+// with the relevant set of changes, end-to-end tests (aka first smoke) have
+// been run agains them, and unit tests have been run on the corresponding
+// code. Here's what we do, some of it in parallel:
 //
-// 1. Prompt the user to do manual testing and either continue or abort.
+// 1. Prompt the user to do manual testing and either continue to "set default"
+//    or abort. Unattended deploys proceed to set default automatically.
 //
-// 2. (Assuming 'continue')  "Prime" GAE to force it to start up
-//    a few thousand instances of the new version we deployed.
+// 2. Tell Google to make our new revision(s) of each service the
+//    default-serving revision (100% traffic allocation).
 //
-// 3. Tell google to make our new version the default-serving version.
+// 3. Start automated monitoring of our Cloud Run and StackDriver logs to check
+//    for an uptick in errors during or after the traffic migration to the new
+//    default.
 //
-// 4. Run end-to-end tests again on now that our new version is default.
-//    This can catch errors that only occur on a khanacdemy.org domain.
+// 4. After all traffic migrations to the new default have completed, run
+//    end-to-end tests again (aka second smoke) now that our new version(s) are
+//    default. This can catch errors that only occur on a khanacdemy.org
+//    domain.
 //
-// 5. Do automated monitoring of our appengine logs to check for an
-//    uptick in errors after the deploy.
+// 5. Prompt the user to either finish up or abort. Unattended deploys will
+//    automatically abort if there were failures and will automatically finish
+//    up if there were not (monitoring failures are ignored due to low
+//    signal-to-noise ratio).
 //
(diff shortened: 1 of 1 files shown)


nathanjd:
@claude review this once.

csilvers:
@claude review once

csilvers:
Looks like claude isn't set up on this repo. :-(  @jeresig how do we add it?

csilvers (changes requested):
> set-default-status being set to "started" AFTER it is set to "finished".

Oh interesting, because the code to set to "started" and the code to set to "finished" run in parallel.

In theory this shouldn't happen because set_default.py shouldn't exit until we're fully migrated, whereas wait_for_default.py should exit when we just start to migrate.  Something weird is definitely going on.  https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27898/execution/node/263/log/ (for set-default) says this:
> 13:16:41  [INFO] Service assign-content already set to 260512-1052-fa5142ebbfc2, exiting.

But https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27898/execution/node/238/log/ (for wait-for-default) says this:
> 13:34:27  INFO:root:Version of assign-content is still 260512-0955-d0e8a60a52e3.
> 13:34:42  INFO:root:Version of assign-content has changed to 260512-1052-fa5142ebbfc2!

wait-for-default hits the `/_api/version` endpoint.  The traffic-directing cookie it depends on hasn't worked in years, so it hits a random instance, it won't definitely exit until the new service is at 100%.  Whereas set-default makes a gcloud API call.  In this case the gcloud API call satisfied this:
```
    if list(existing_split.keys()) == [version]:
```
which sure seems like it could only happen if 100% of the traffic was on the new version, so I don't know how set-default could think we're already at fa5142 whereas wait-for-default does not.  I also don't understand _why_ the service would already be set to the new version (in set-default): what would have done that?  A different set-default?  Something isn't adding up here.

I'm hesitant to land this, because clearly, in this instance, set_default.py exited before assign-content was actually getting 100% of the traffic.  I think before we do something like this, we'd want to modify set_default.py itself to call wait_for_version_change (from wait_for_default.py) before exiting, to make sure that not only does gcloud think the traffic has migrated, but our API is reporting that as well.

But I wonder if we should keep things as they are, and instead add code here that sends to slack when set-default.py finishes after wait-for-default.py (maybe significantly after, like at least a few seconds).  And then we can jump in and see what gcloud thinks the traffic distribution is like, and what our api call thinks it's like, and see if that gives more insight.

@nathanjd is this the one where you had to manually edit the yaml for it to complete?  I wonder if the issue is that google is setting traffic to 100% for the API call, but not actually migrating anything because the yaml is invalid.

nathanjd:
@csilvers 

> is this the one where you had to manually edit the yaml for it to complete? I wonder if the issue is that google is setting traffic to 100% for the API call, but not actually migrating anything because the yaml is invalid.

This one I linked specifically in the PR description, yes. But the pattern is pretty consistent and @MiguelCastillo mentioned he hits it with every deploy. I can add links to the logs for the other instances as well. I've responded to many @deploy-support-backend requests that hit this issue (no finish up prompt) without any hanging update-traffic calls. Today's deploy with update-traffic timing out against assign-content was unique. Though I have been seeing update-traffic timeouts more often recently which is very disappointing and concerning given I expected the mutex to fix those issues.

I'll look into the rest!

edit - I tried to retrieve the previous instances of this but the console logs are gone despite me having clicked "Keep this build forever". I guess I'll have to start downloading the logs and uploading them to the tickets. I'll poke through the jobs that do still have logs to see if I can find more examples.

Occurrences of "started" after "finsished".

- https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27892/consoleFull (this one was just `fastly-khanacademy-compute` which could explain why @MiguelCastillo has been hitting this so consistently.)
- https://jenkins.khanacademy.org/job/deploy/job/deploy-webapp/27867/consoleFull (also just `fastly-khanacademy-compute`)

csilvers:
Ah, in these cases what happens is that wait-for-default waits 15 seconds between checks, so it's quite plausible that set-default will finish and then wait-for-default won't notice things are done for another 10 seconds or so.  That is probably the common case.

csilvers (approved):
OK, both of the other examples you showed are due to wait-for-default having a long delay between polls.  I can believe that's the common case.  I still suspect bad things will happen when the yaml is corrupted, and it would make sense to have set-default.py explicitly check `/_api/version` instead of (or in addition to) relying on what gcloud says, but that doesn't need to happen first.  But it would be good if you could take it on at some point.  With a comment like:
```
// Apparently Google can say traffic is 100% to the new version but not actually
// route the traffic if the cloudrun yaml file is corrupted.  So let's make sure it's
// actually routed, and if not give an error.
<code to check /_api/version>
```

That is nice because we could make the error point out the possibility of the yaml being bad, which is otherwise hard to diagnose.

nathanjd:
I'll look into the issues we saw with the `update-traffic` timeouts we saw for the `assign-content` service in this ticket:

https://khanacademy.atlassian.net/browse/INFRA-11056

PR #391, opened 2026-05-19
Send a tag for make-allcheck so it uses github actions for webapp_test

jaredly:
## Summary:
required to let it use github actions.
defaults to master, same as GIT_REVISION.

Issue: none

## Test plan:
🤞

### DIFF
--- a/jobs/make-allcheck.groovy
+++ b/jobs/make-allcheck.groovy
@@ -34,6 +34,7 @@ def runAllTests() {
     build(job: '../deploy/webapp-test',
           parameters: [
              string(name: 'GIT_REVISION', value: params.GIT_REVISION),
+             string(name: 'GIT_TAG', value: "master"),
              string(name: 'BASE_REVISION', value: ""),
              string(name: 'SLACK_CHANNEL', value: "#backend"),
              booleanParam(name: 'FORCE', value: params.FORCE),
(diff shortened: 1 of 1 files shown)


jobs/make-allcheck.groovy
```diff
@@ -10,6 +10,12 @@ import org.khanacademy.Setup;
 
 new Setup(steps
 
+).addStringParam(
+   "GIT_TAG",
+   """The git tag or branch name to be used to select the github actions workflow.
+Defaults to master if no tag is provided.""",
```
csilvers:
```suggestion
If GIT_REVISION is a sha, leave this blank.""",
```

csilvers (approved):
Thank you!

nathanjd (approved):

jobs/make-allcheck.groovy
jaredly: hmmmmmm I find this misleading. if GIT_REVISION is a sha, there may well be a branch they want to run the webapp-test workflow off of that's not master.

csilvers: I think if they want to run off a branch, they set GIT_REVISION to the branch.  In my experience, you only set GIT_REVISION to a sha if you want to test something that is not a branch or tag.

PR #392, opened 2026-05-20
Migrate merge-branches to optional GitHub Actions bridge

jaredly:
## Summary:
- Adds an opt-in `USE_GITHUB_BRIDGE` path to `merge-branches.groovy`, preserving the existing Jenkins merge behavior as the default.
- Adds a `merge-branches` GitHub Actions workflow in `webapp` plus a Python merge script that preserves Jenkins merge semantics.
- Keeps Slack and Buildmaster notification ownership in Jenkins during bridge mode by downloading a workflow result artifact and calling `buildmaster.notifyMergeResult(...)` from Jenkins.
- Extends `runGithubAction` with `dispatchAndWait(...)` so callers can get the completed Actions run id.

Issue: https://khanacademy.atlassian.net/browse/INFRA-11001

## Test plan:
- `python3 -m unittest deploy.merge_branches_test`
- `python3 -m py_compile deploy/merge_branches.py deploy/merge_branches_test.py`
- `actionlint .github/workflows/merge-branches.yml`
- `git diff --check`
- Local Groovy parse check for changed Jenkins scripts, with Jenkins `@Library` stripped for local parsing.

### DIFF
--- a/jobs/merge-branches.groovy
+++ b/jobs/merge-branches.groovy
@@ -8,11 +8,14 @@
 @Library("kautils")
 // Classes we use, under jenkins-jobs/src/.
 import org.khanacademy.Setup;
+import groovy.json.JsonSlurper;
 // Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
 //import vars.buildmaster
 //import vars.exec
 //import vars.kaGit
 //import vars.notify
+//import vars.runGithubAction
+//import vars.withSecrets
 //import vars.withTimeout
 
 new Setup(steps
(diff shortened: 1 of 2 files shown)


jobs/merge-branches.groovy
```diff
@@ -89,21 +97,72 @@ String getGaeVersionName() {
...
+
+
+def githubMergeResult(String runId) {
+   String resultDir = "merge-branches-github-result-${env.BUILD_NUMBER}";
+   dir(resultDir) {
+      deleteDir();
+   }
+
+   String token = withSecrets.getGithubActionsToken();
+   withEnv(["GITHUB_TOKEN=${token}"]) {
```
csilvers, line 118: Sad that this seems to be the only way to do this.

Is there a way to get this data via curl, like you do in runGithubAction.groovy?

csilvers (approved):
Looks good to me!

jobs/merge-branches.groovy
jaredly: there's not a simple curl that would do the same thing, it would take several steps. Is withEnv + a single exec more dangerous than I think it is? I would expect it to only expose the env vbl to the single shell execution

csilvers: Env vars show up in `ps` so anyone -- or any other process -- who is on the machine can see it.  It's not the end of the world -- we store passwords in plaintext files for our jenkins jobs, and this is not worse than that.  But I was hoping to make life better.  I don't think you need to change this if it's not simple.

PR #393, opened 2026-05-22
Add GitHub bridge to build-webapp

nathanjd:
## Summary:
As part of porting our deploy jobs from jenkins to GitHub Actions, add a
param `USE_GITHUB_BRIDGE` to call the build-webapp.yml GitHub workflow
from the jenkins job instead of doing the work on jenkins runners.

Add `GIT_TAG` parameter as it is needed by GitHub workflows. Buildmaster
will be updated to additionally pass this.

Issue: https://khanacademy.atlassian.net/browse/INFRA-10995

Test plan:

1. Use replay to test that the non-github code path still works before
   landing this PR.
   https://jenkins.khanacademy.org/job/deploy/job/build-webapp/75331/
2. After landing, monitor build-webapp job for failures.
3. Once build-webapp.yml has been deployed to webapp, test its
   invocation by manually triggering build-webapp with
   `USE_GITHUB_BRIDGE=true`.

### DIFF
--- a/jobs/build-webapp.groovy
+++ b/jobs/build-webapp.groovy
@@ -567,11 +567,6 @@ def finishWithFailure(why) {
 }
 
 def run(Boolean useGithub) {
-   // TODO(ebrown): Remove: onWorker logs, so not needed here too
-   notify.log("Starting ${env.JOB_NAME} " +
-              "${params.REVISION_DESCRIPTION} ${env.BUILD_NUMBER}", [
-   ]);
-
    notify([slack: [channel: params.SLACK_CHANNEL,
                    sender: 'Mr Monkey',
                    emoji: ':monkey_face:',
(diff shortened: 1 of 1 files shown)


jobs/build-webapp.groovy
```diff
@@ -1,27 +1,29 @@
-// Upload a new version of (backend) webapp code to App Engine and/or Cloud Run.
+// Upload a new version of (backend) webapp code to Cloud Run.
```
csilvers, line 1: Thanks for fixing up the comments!

jobs/build-webapp.groovy
```diff
@@ -560,10 +566,12 @@ def finishWithFailure(why) {
    }
 }
 
+def run(Boolean useGithub) {
+   // TODO(ebrown): Remove: onWorker logs, so not needed here too
+   notify.log("Starting ${env.JOB_NAME} " +
+              "${params.REVISION_DESCRIPTION} ${env.BUILD_NUMBER}", [
+   ]);
```
csilvers: Where did this come from?  I don't see it in the old version.  And agree we don't need to log it.

csilvers (approved):
Great!

jobs/build-webapp.groovy
nathanjd: I copied it from our ported webapp-test job. I'll remove it.

PR #394, opened 2026-05-26
Bridge delete-versions job to github actions

jaredly:
## Summary:
following https://khanacademy.atlassian.net/wiki/spaces/INFRA/pages/4712169576/Webapp+Deploys+Jenkins+-+GitHub+Project+Plan

webapp side https://github.com/Khan/webapp/pull/39676

Issue: https://khanacademy.atlassian.net/browse/INFRA-10996

## Test plan:
🤞 the workflow has a boolean that will allow us to test it out

### DIFF
--- a/jobs/delete-version.groovy
+++ b/jobs/delete-version.groovy
@@ -8,6 +8,7 @@ import groovy.json.JsonSlurperClassic;
 //import vars.exec
 //import vars.kaGit
 //import vars.notify
+//import vars.runGithubAction
 //import vars.withTimeout
 //import vars.withVirtualenv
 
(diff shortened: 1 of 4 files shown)


jobs/delete-version.groovy
```diff
@@ -78,44 +89,78 @@ def deleteVersion() {
...
+         }
+      }
+   }
+}
+
+def runInGithub() {
+   stage("Deleting") {
+      // Keep the Jenkins lock during bridge execution so this job still
+      // coordinates with deploy-webapp while traffic migration remains in
+      // Jenkins.
```
csilvers, line 135: 👍

jobs/delete-version.groovy
csilvers, line 85: It would be good to handle params.DRY_RUN here too, since you're adding it.

csilvers (approved):
Looks good (but see DRY_RUN comment below)

PR #396, opened 2026-06-02
Fix build-current-sqlite to enable copying files to gcs again.

csilvers:
## Summary:
This script broke when I fixed the datastore-emulator container to no
longer mount gcloud credentials, which it doesn't need.  (Actually, as
part of a different security audit, I changed it to mount them
read-only, which made gcloud reject them for some reason.  Then, when
I realized what was going on, I removed the mounting entirely,
realizing it wasn't necessary anymore.)  But it turns out we were
depending on that capability in this script, to copy files from the
container to gcs.

I enable that functionality a different way now.  I manually copy the
credential needed into the countainer, since you can't mount volumes
when running `docker exec` and I wouldn't want to mount the
credentials volume read-write anyway.  And then I use that account to
do the copying, which uses gcloud now instead of gsutil.  And all is
well.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1780329614592999

## Test plan:
I used `replay` to run this new script, at TODO.

### DIFF
--- a/jobs/build-current-sqlite.groovy
+++ b/jobs/build-current-sqlite.groovy
@@ -7,6 +7,7 @@
 // Classes we use, under jenkins-jobs/src/.
 import org.khanacademy.Setup;
 // Vars we use, under jenkins-jobs/vars/.  This is just for documentation.
+//import vars.exec
 //import vars.kaGit
 //import vars.notify
 //import vars.onWorker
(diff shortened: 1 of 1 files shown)


dnerdy (approved):
Seems like a good strategy to me!

What is `replay`, btw?

dbraley (approved):
Ah, ok, so this doesn't fundamentally change how this process works at all, just makes it so the normal runtime doesn't need those permissions at all, and this one instance that does is the only one that gets them. Cool :shipit:

dbraley:
> Seems like a good strategy to me!
> 
> What is `replay`, btw?

Replay lets you rerun a previous job/build/whatever with the exact same parameters but different code. It's a little wonky because you have to copy-paste the code into the UI rather than testing it on a git branch, but it works fine and as far as I'm aware is the only or at least easiest way to test small changes to jenkins-jobs.

dnerdy:
> > Seems like a good strategy to me!
> > What is `replay`, btw?
> 
> Replay lets you rerun a previous job/build/whatever with the exact same parameters but different code. It's a little wonky because you have to copy-paste the code into the UI rather than testing it on a git branch, but it works fine and as far as I'm aware is the only or at least easiest way to test small changes to jenkins-jobs.

I suppose the main thing I'm wondering is: how does one run `replay`?

csilvers:
> I suppose the main thing I'm wondering is: how does one run replay?

It's one of the options in the jenkins menu for a job run:

image

dnerdy:
> > I suppose the main thing I'm wondering is: how does one run replay?
> 
> It's one of the options in the jenkins menu for a job run:

Ah, okay. Thanks for the screenshot!

PR #397, opened 2026-06-02
Prune `znd-*` tags the same way we prune `buildmaster-*` tags.

csilvers:
## Summary:
We have more than 4000 of these, so they're overdue for pruning!

Issue: none

## Test plan:
I ran the `git tag | grep` command manually and the output looked correct.

### DIFF
--- a/weekly-maintenance-jobs.sh
+++ b/weekly-maintenance-jobs.sh
@@ -82,6 +82,38 @@ clean_invalid_branches() {
 # The merge-branches jenkins job creates a git-tag to make sure the
 # merged code isn't gc-ed by Jenkins.  It only needs to survive for
 # the length of a deploy: maybe an hour or so.  We keep these tags
+# around a week.  Same goes for the deploy-znd jenkins job.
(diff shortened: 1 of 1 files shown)


weekly-maintenance-jobs.sh
```diff
@@ -82,6 +82,38 @@ clean_invalid_branches() {
 # The merge-branches jenkins job creates a git-tag to make sure the
 # merged code isn't gc-ed by Jenkins.  It only needs to survive for
 # the length of a deploy: maybe an hour or so.  We keep these tags
+# around a week.  Same goes for the deploy-znd jenkins job.
+prune_buildmaster_and_znd_tags() {
```
nathanjd, line 86: This seems to duplicate the `prune_buildmaster_tags() {` function below. Should that one be removed?

csilvers: Oh oops, that was a cut-and-paste error on my part!

nathanjd (approved):
Looks good!

PR #398, opened 2026-06-04
Bridge deploy-znd job to GitHub Actions (INFRA-10999)

jaredly:
Ticket: https://khanacademy.atlassian.net/browse/INFRA-10999

## Summary

- Adds a `USE_GITHUB_BRIDGE` boolean param to the `deploy-znd` Jenkins job
- When true, resolves the git revision to a SHA and dispatches to the new `deploy-znd.yml` GitHub Actions workflow via `runGithubAction`, passing through all relevant inputs (`git_revision`, `version`, `services`, `slack_channel`, `slack_thread`, `actor`)
- When false (default), falls back to the existing Jenkins deploy path unchanged
- Companion PR in webapp: https://github.com/Khan/webapp/pull/39788

## Test plan

- [ ] Verify existing Jenkins path still works with `USE_GITHUB_BRIDGE=false`
- [ ] Trigger a ZND with `USE_GITHUB_BRIDGE=true` and confirm the GitHub Actions workflow runs and posts Slack notifications correctly

### DIFF
--- a/jobs/deploy-znd.groovy
+++ b/jobs/deploy-znd.groovy
@@ -463,7 +463,11 @@ def run(Boolean useGithub) {
 // just to wait on GHA). Otherwise use a separate worker type, identical to
 // build-worker, so znds don't make a mess of our build caches for the main deploy.
 if (params.USE_GITHUB_BRIDGE) {
-   onMaster('3h') { run(true) }
+   onMaster('3h') {
+      run(true)
+   }
 } else {
-   onWorker('znd-worker', '3h') { run(false) }
+   onWorker('znd-worker', '3h') {
+      run(false)
+   }
 }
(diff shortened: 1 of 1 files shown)


jobs/deploy-znd.groovy
```diff
@@ -404,23 +411,59 @@ def deploy() {
...
          }
       }
    }
 }
+
+// When USE_GITHUB_BRIDGE=true, run on master (no need to spin up a znd-worker
+// just to wait on GHA). Otherwise use a separate worker type, identical to
+// build-worker, so znds don't make a mess of our build caches for the main deploy.
+if (params.USE_GITHUB_BRIDGE) {
+   onMaster('3h') { run(true) }
```
csilvers: Nit: I think in the other jobs you expand this out to 3 lines.  If so I'd do the same here for consistency.

csilvers (approved):
Seems right!

PR #402, opened 2026-06-24
Give buildmaster up to 1 minute to restart before giving up

nathanjd:
## Summary:
Previously the retry block had no delay between attempts, so all
retries exhausted within seconds — not enough time for a buildmaster
instance to finish starting up. Now sleeps 20s between each of 4
attempts, giving it ~60s to recover.

Issue: https://khanacademy.atlassian.net/browse/INFRA-11111

Test plan:

1. Use replay feature to run a merge-branches job to ensure there are
   no breaking code changes.
2. Monitor all jenkins jobs after landing.


csilvers (approved):
Good idea!

PR #404, opened 2026-07-07
Fix merge-branches github dispatch error

jaredly:
## Summary:
The github workflow doesn't expect `buildmaster_deploy_id`, and so github was refusing to start the workflow.

additionally, we were suppressing all output from a curl command that was failing 😭 so I had to do some manual testing to figure out what was wrong. https://jenkins.khanacademy.org/job/deploy/job/merge-branches/370517/execution/node/53/log/

I also notice that the auth token was being shown in plain text 😩 so I fixed that too

Issue: FEI-XXXX

## Test plan:
run it again with these inputs https://jenkins.khanacademy.org/job/deploy/job/merge-branches/370517/parameters/?

### DIFF
--- a/jobs/merge-branches.groovy
+++ b/jobs/merge-branches.groovy
@@ -139,7 +139,6 @@ def runInGithub() {
          git_revisions: params.GIT_REVISIONS,
          commit_id: params.COMMIT_ID,
          revision_description: params.REVISION_DESCRIPTION,
-         buildmaster_deploy_id: params.BUILDMASTER_DEPLOY_ID,
       ]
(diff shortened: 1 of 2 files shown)


csilvers (approved):
Thank you for fixing the leaked token!  Your (or claude's) groovy-fu exceeds my own.

Do we need to rotate the token now?  Eventually those old logs will get reaped, and we could also do it earlier if we wanted.

nathanjd (approved):

jaredly:
yeah we ought to rotate the token

PR #405, opened 2026-07-07
[🔥AUDIT🔥] Fix json parsing in merge-branches

jaredly:
🖍 _This is an audit!_ 🖍

## Summary:
lol groovy, what are you even.

> Root cause: jobs/merge-branches.groovy parsed the downloaded result JSON with `JsonSlurper`, which returns a `LazyMap`. Jenkins CPS can’t serialize that object while pipeline steps are still running. The catch block then reported the buildmaster commit as failed with no SHA, so the success was immediately overwritten by the failure report.

Issue: none

## Test plan:
I did a "replay" with this change, and it worked

### DIFF
--- a/jobs/merge-branches.groovy
+++ b/jobs/merge-branches.groovy
@@ -8,7 +8,7 @@
 @Library("kautils")
 // Classes we use, under jenkins-jobs/src/.
 import org.khanacademy.Setup;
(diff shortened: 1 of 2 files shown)


csilvers (commented):
Yeah, none of that fancy _new_ stuff!

PR #407, opened 2026-07-29
Match GHA runs by unique dispatch_id instead of head_sha

nathanjd:
## Summary:
workflow_dispatch doesn't return a run ID synchronously, so runGithubAction
has to poll and guess which run is "ours" afterward. Matching on
head_sha + head_branch is ambiguous when a concurrent or still-running
prior dispatch shares the same ref and SHA (e.g. concurrent
merge-branches builds racing against an unmoved master), which caused
Jenkins build 373960 to watch and report success for the wrong run.

Generate a unique dispatch_id per call and send it as a workflow input;
match the polled run by display_title (which GitHub populates from a
run-name: template) instead. This requires the corresponding Khan/webapp
workflows to declare a dispatch_id input and run-name: template before
this lands, since GitHub rejects undeclared workflow_dispatch inputs.

Dependent on https://github.com/Khan/webapp/pull/41171

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

Issue: https://khanacademy.atlassian.net/browse/INFRA-11152

Test plan:

As no jobs currently defaults to `USE_GITHUB_BRIDGE=true` at the moment,
it is safe to test each job with `USE_GITHUB_BRIDGE=true` after landing
this.

### DIFF
--- a/vars/runGithubAction.groovy
+++ b/vars/runGithubAction.groovy
@@ -15,14 +15,24 @@ def _githubApiHeaders(String token) {
 //   repo     - GitHub repo, e.g. "Khan/webapp"
 //   workflow - Workflow filename, e.g. "webapp-test.yml"
 //   ref      - Branch name to run the workflow on
-//   headSha  - Resolved SHA for the ref (used to identify the new run)
 //   inputs   - Map of string→string workflow inputs (optional)
 //   token    - a github token, used to perform the workflow_dispatch
 //
-// Dispatches the workflow via the GitHub API, then polls for up to 30s
-// to locate the new run and return its ID.
+// Dispatches the workflow via the GitHub API, tagging the dispatch with a
+// unique dispatch_id input, then polls for up to 30s to locate the new run
+// and return its ID.
 def _dispatch(Map args) {
-    def payload = JsonOutput.toJson([ref: args.ref, inputs: args.inputs ?: [:]])
+    // A unique-per-call ID we can use to unambiguously identify our run once
+    // it's created. workflow_dispatch doesn't return a run ID synchronously,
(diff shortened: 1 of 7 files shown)


vars/runGithubAction.groovy
```diff
@@ -37,20 +47,18 @@ def _dispatch(Map args) {
...
         sleep(3)
         def response = httpRequest(
             customHeaders: _githubApiHeaders(args.token),
             httpMode: "GET",
-            url: "https://api.github.com/repos/${args.repo}/actions/runs?event=workflow_dispatch&per_page=10${shaFilter}")
+            url: "https://api.github.com/repos/${args.repo}/actions/workflows/${args.workflow}/runs?event=workflow_dispatch&per_page=10")
         def runs = new JsonSlurperClassic().parseText(response.content)
         for (run in runs.workflow_runs) {
-            if (run.head_branch == args.ref) {
+            if (run.display_title == dispatchId) {
```
csilvers, line 61: Does this work?  I don't know how display_title is computed, but in https://github.com/Khan/webapp/pull/41171 you do things like:
```
run-name: ${{ inputs.dispatch_id != '' && format('merge-branches {0}', inputs.dispatch_id) || 'merge-branches' }}
```
Is `run-name` what turns into the display-title?  Because it has other text in it besides the dispatch-id.

csilvers (commented):
Again, I don't feel really competent to review this.  If you need me as a reviewer, maybe it'll be easiest to go over it in person next week.

vars/runGithubAction.groovy
nathanjd: Switched to a contains check.

PR #408, opened 2026-08-17
Add ONLY_POST_ON_ERROR to webapp-test, and use it from make-allcheck.

csilvers:
## Summary:
Normally, when running the webapp-test github action, we post to slack
with the results.  But there's been a request to not do so in the
(hopefully!) common case that the nightly "allcheck" run passes.  We
only want to see a message if the allcheck run unexpectedly fails.

In Khan/webapp#41541, I added support for a new `--only-post-on-error`
flag in webapp.  In this PR, I wire that thread through to the
make-allcheck job in jenkins.

Issue: https://khanacademy.slack.com/archives/C0A0EP14K/p1786353644754299

## Test plan:
I ran:
```
groovy jobs/make-allcheck.groovy
```
to check for typos.  But mostly, fingers crossed!


nathanjd (approved):
Looks good. Just remember the workflow needs to be merged before this is so that the GH workflow doesn't throw an "unknown parameter" error.

